PR #211, opened 2015-04-24
Add standardized dissolution model checking

### LINKED ISSUE #28: Enforce check that dissolution formula embedded in formation
Really only relevant once dissolution formula becomes more flexible

smjenness:
Closes #28


ebey:
I'm not totally familiar with testthat, but I believe the build may have failed because the expected error in the last test for diss_check does not match the actual error the function returns.

https://github.com/statnet/EpiModel/blob/f1a7c26ca2db7998604797393328a5ac5d999bf8/tests/testthat/test-netest.R#L81

https://github.com/statnet/EpiModel/blob/f1a7c26ca2db7998604797393328a5ac5d999bf8/R/netest.R#L314-315

smjenness:
No, that test is fine (it only needs a partial match on the expression). The problem is in the first test in test-netoffset.r

smjenness:
Nice use of `regmatch`. Thanks!

Issue #230, opened 2015-09-18
Dendogram/phylo trees for transmission chains from epidemic simulations


smjenness:
@skyebend Here is a simple test case for getting started. Thanks for your help!

```r
library(EpiModel)
nw <- network.initialize(n = 100, directed = FALSE)
formation <- ~edges
target.stats <- 50
coef.diss <- dissolution_coefs(dissolution = ~offset(edges), duration = 20)
est1 <- netest(nw, formation, target.stats, coef.diss, verbose = FALSE)

param <- param.net(inf.prob = 0.5)
init <- init.net(i.num = 1)
control <- control.net(type = "SI", nsteps = 40, nsims = 1, verbose.int = 0, use.pids = FALSE)
mod1 <- netsim(est1, param, init, control)

tm <- get_transmat(mod1)
tm
```

The parent for any tree branch would be the ID in `inf` and the child would be the id in `sus`.

skyebend:
Ouch, this is much harder than anticipated, took me 3 days!   The plot.phylo code is pretty fragile and seems to depend on lots of undocumented elements of how it parses trees from its textual input syntax, crashes R on any errors.  Sort of have it figured out to at least render a tree.

https://github.com/statnet/EpiModel/tree/dendo_plot

Some things I've learned:  

All of the 'tips' (end vertices) must be numbered starting from 1.  The 'nodes' (vertices with children) must have ids > tips, and the 'root' must have id = length(tips)+1.  This requires walking through the tree to relabel the internal phylo node ids. 

Problems:

1) phylo representation  prohibits nodes that don't split (i.e. they have only one child).  This means that a linear sequence of infections can't be shown (A->B->C) and they happen a lot.  This can be avoided by adding a `collapse.singles()` call, but that seems to scramble the names. Possibly we can solve this by adding some dummy tips?.  Or I can try to set it up to merge the two elements and give them a combined name?

2) still need to correctly handle common cases of multiple trees

3) I'd like to add a 'transmat' class to the transmat object so we can do S3 dispatch

4) add timing info

5) obviously docs and tests and such

smjenness:
Thanks, Skye. I understand now why I was getting frustrated in doing this after an hour. I'm not sure the best way to handle linear sequences. Let me get back to you on that one.

Feel free to add a transmat class, probably should go here:
https://github.com/statnet/EpiModel/blob/master/R/saveout.R#L158-L165

skyebend:
[...]

PR #232, opened 2015-10-15
transmat & dendogram plot 

skyebend:
This is an initial version of the transmat class and dendogram plotting (#230) with basic functionality.  requires updated version of ndtv (on CRAN now).  Includes basic docs and tests. We could do additional configuration of preset colors etc of the various plot types to match EpiModel.  Passes checks on my machine, but branch doesn't pass CI (It appears to be just because of the ergm/network import warnings)


smjenness:
Thanks, Syke. Travis is still erroring because of the redundant as.edgelist import from the CRAN version of ergm. I'll review and merge from here. Thanks for all your work on this!

smjenness:
@skyebend I made some cosmetic code changes last night, but now I am getting the following error when I try to reoxygenize and rebuild the package. Any idea how to fix this? It is pushing the limits of my knowledge of S3 classes! Thank you.

```
Error: object 'as.phylo' not found whilst loading namespace 'EpiModel'
```

skyebend:
Perhaps my roxygen wasn't working quite right and was not generating the offending S3method definition in the namespace?

skyebend:
"whilst" :: ye olde CRAN warnings? :-)

smjenness:
Ok, I guess let's go for it and move ape to an import. I think you can just import the core couple of functions needed, as I do for the other packages in the meta doc:
https://github.com/statnet/EpiModel/blob/master/R/EpiModel-package.r

skyebend:
I'm wondering if something is funky with the ape package namespace, as I can't get Import to work as expected.

smjenness:
@skyebend I got things working (I think roxygen was stuck, so I had to do a little manual editing of the NAMESPACE). Also, I had do disable the graphviz option on plot.transmat as that does not work on my computer. I think this is ready to merge now unless you wanted to make any final changes...

skyebend:
The CRAN checks are generating NOTES about the missing S3 method registration, so I don't think they will accept.  
```
checking S3 generic/method consistency ... NOTE

Found the following apparent S3 methods exported but not registered:

  as.phylo.transmat

See section ‘Registering S3 methods’ in the ‘Writing R Extensions’
```

  If we remove the gv_tree option, should also remove mention in docs text.  

Also
```
* checking DESCRIPTION meta-information ... NOTE

Package listed in more than one of Depends, Imports, Suggests, Enhances:

  ‘ape’
```

And think we have to leave in the  requireNamespace("ndtv") before the transmission tree call, unless ndtv is moved from Suggests to Depends because we don't know for sure that it is loaded

These should all be easy, so I'll go ahead and do them

skyebend:
Seems like the intention of your edits was to move ape from Depends to Imports? But accidentally left in both?.  But when I remove it from Depends, the checks fail again because as.phylo is not found.  call/skype?

skyebend:
OK, I think that is ready for merge.  Currently, if ape is loaded, as.phylo(tm) works, otherwise you have to call as.phylo.transmat(tm).   Obviously we can do much more work to make the phylo plots pretty and informative, but should probably be separate tickets for future release

smjenness:
Looks good to me, thanks for this great start!

skyebend:
not sure why I didn't find this earlier, but the `epinet` package has really nice `plotepitree` function that also indicates SIR status

PR #237, opened 2016-02-19
as.phylo.transmat builds the phylo tree with vertices as tips

skyebend:
as per discussion with Joshua Herbeck


smjenness:
Thanks, Skye!

PR #238, opened 2016-03-01
added optional death times to as.phylo.transmat

skyebend:


skyebend:
Maybe there is a better name for the parameter.   vertex.death.times   or vertex.terminations  might be clearer?

smjenness:
I think `exit.times` will be more appropriate, because nodes may leave for death or other reasons

PR #241, opened 2016-04-12
additional phylo tree updates

skyebend:
* fixes a bug where vertex.exit.times were not correctly translated to phylo edge durations for some transmission matrices
* added support for transmission matrices with multiple seeds (a list of phylo objects, one for each seed is returned)


smjenness:
Thanks, Skye!

PR #247, opened 2016-07-29
Readthru

kweiss2:


R/dendo.R
```diff
@@ -232,13 +232,13 @@ as.network.transmat <- function(x, ...){
...
 #' @description Plots the infection tree described in a \code{\link{transmat}}
-#'              object in one of several styles: phylogentic tree, an un-rooted
+#'              object in one of several styles: phylogenetic tree, an un-rooted
 #'              network, a hierarchical tree, or a transmissionTimeline.
 #'
-#' @param x a \code{\link{transmat}} object to be plotted
-#' @param style character name of plot style. One of "phylo", "network",
+#' @param x A \code{\link{transmat}} object to be plotted
+#' @param Style character name of plot style. One of "phylo", "network",
 #'        or "transmissionTimeline"
```
smjenness: I think you meant to capitalize Character not Style (capitalization of params must be the same as in the function)

kweiss2: Good catch, you are correct!

PR #250, opened 2016-09-30
Fast edgelist

skyebend:
I set the build test to require the new version of tergmLite that includes the terms, so it will fail until the PR for tergm lite is approved.


smjenness:
Closing pull request as fast_edgelist code will remain in branch and not be merged into master, per our earlier discussion about CRAN-related NAMESPACE problems.

Issue #252, opened 2017-04-07
Add grid = TRUE/FALSE toggle for all plots


smjenness:
This would go in the arguments of all the plot functions within EpiModel. If `grid = TRUE` then after plotting the plot but before adding the legend, we would add:

```r
grid()
```

kweiss2:
@smjenness Do you want this option added to the tutorials as well (e.g. functions like renderPlot outside of the plot.R file?

kweiss2:
They show up in the shiny app - server files

smjenness:
Oh right -- no you don't need to have toggles for this in the Shiny apps

PR #254, opened 2017-05-05
Grid toggle for #252 (plot.dcm, plot.icm, plot.netsim)

kweiss2:


smjenness:
Thanks!

1. Set the default to grid = FALSE
2. Add examples to each of the places where it is used (it looks like you did this for one function but not the others)
3. Can you add a unit test for this to check for this (we can discuss how to do that)

smjenness:
Also:

4. Wrap the roxygen text at about 80 characters
5. Add a grid option for plot.netdx too

kweiss2:
Did you this option in each of the examples?

smjenness:
Yes, examples in each of the functions in which you added the option

smjenness:
Thanks @kweiss2

Issue #276, opened 2018-08-29
Error message when 1/duration & d.rate too close


ejanderson1:
When the rate of dissolution is too close to d.rate the error message reads:

```ln log(pg/(ps2-pg)): NaNs produced```

Needs to be updated to something more informative.

**Comment by smjenness** [comment]
This combo still results in an error:

```r
dissolution_coefs(~offset(edges), 60, 0.01)
```

smjenness:
@andsv2 is currently working on this.

andsv2:
> This combo still results in an error:
> 
> ```r
> dissolution_coefs(~offset(edges), 60, 0.01)
> ```

Apologies if this is a simple question, but is the above meant to work? The operand (60, 0.01) resolves to less than 0 and shouldn't work.

Also is there a point at which the adjusted coef. of dissolution is unreasonable? For example if coef.diss = 3 and adj. coef.diss = K >>3.

smjenness:
No, the above should not work. Anything that currently produces the warning message:

```r
Warning message:
In log(pg/(ps2 - pg)) : NaNs produced
```

because the adjusted coefficient is `NaN` should error instead.

The adjusted coefficient may be much greater than the crude coefficient. For example, this is fine:

```r
> dissolution_coefs(~offset(edges), 50, 0.01)
Dissolution Coefficients
=======================
Dissolution Model: ~offset(edges)
Target Statistics: 50
Crude Coefficient: 3.89182
Mortality/Exit Rate: 0.01
Adjusted Coefficient: 9.190138
```

smjenness:
Closed with #287

PR #287, opened 2018-10-24
dissolution_coefs function update

andsv2:
#276 

Updated logic test to account for individual survival rate being too close to edge duration such that operand resolves negative.

Also added more informative error message.


R/net.utils.R
```diff
@@ -981,35 +976,38 @@ get_degree <- function(x) {
...
   # epi
   x$epi <- lapply(x$epi, function(r) r[rows, ])
-
+  
   # control settings
   x$control$start <- 1
   x$control$nsteps <- max(seq_along(rows))
-
+  
   return(x)
```
smjenness: All these additions of whitespace need to be reverted back. Might be a tab option you have set in Rstudio that generated all of them.

PR #288, opened 2018-10-25
Make plot.netsim formation plots consistent with plot.netdx

andsv2:
Changed plot.netsim to match CVM plot.netdx code.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -2198,14 +2198,77 @@ plot.netsim <- function(x, type = "epi", y, popfrac = FALSE, sim.lines = FALSE, 
     }
 
     # Default colors
+   # if (missing(sim.col)) {
+  #    if (nstats > 8) {
+    #    sim.col <- brewer_ramp(nstats, "Set1")
+    #  } else {
+    #    sim.col <- brewer.pal(9, "Set1")[1:(nstats + 1)]
+    #    if (nstats >= 6) {
+    #      sim.col <- sim.col[-which(sim.col == "#FFFF33")]
+    #    }
+    #  }
+    #}
+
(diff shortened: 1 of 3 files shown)


R/plot.R
```diff
@@ -2198,15 +2198,78 @@ plot.netsim <- function(x, type = "epi", y, popfrac = FALSE, sim.lines = FALSE,
...
+  #    if (nstats > 8) {
+    #    sim.col <- brewer_ramp(nstats, "Set1")
+    #  } else {
+    #    sim.col <- brewer.pal(9, "Set1")[1:(nstats + 1)]
+    #    if (nstats >= 6) {
+    #      sim.col <- sim.col[-which(sim.col == "#FFFF33")]
+    #    }
+    #  }
+    #}
+
```
smjenness: This unused code block can now be removed

smjenness:
@andsv2 I'm really not sure why the first check is not passing. I will still merge in after you make the change above plus:

Can you and @connor-vanmeter also set the default color for single output stat plots to the same blue color as the `plots.joined=FALSE` plots. For example, this should also be that blue color:

```r
plot(mod, type = "formation", stats = "meandeg",
     sim.lwd = 2)
```

but it is currently red.

connor-vanmeter:
The color was defaulting to red due to it being the first color from `brewer_ramp(nstats,"Set1")` - "#E41A1C".  When plot colors (mean.col, sim.col, qnts.col, targ.col) are not specified, they take on the default values for sim.col - "Set1" palette colors.
@andsv2 is looking into setting the default color to the dodgerblue3 when nstats==1 is TRUE.

PR #295, opened 2018-11-14
Param.R helper messages

andsv2:


R/param.R
```diff
@@ -135,6 +135,16 @@ param.dcm <- function(inf.prob, inter.eff, inter.start, act.rate, rec.rate,
...
+  if ("b.rate" %in% names.dot.args) {
+    a.rate <- dot.args$b.rate
+    message("EpiModel 1.7.0 onward renamed the birth rate parameter b.rate to a.rate. See documentation for details.")
+  }
+
+  if ("b.flow" %in% names.dot.args) {
+    a.flow <- dot.args$b.flow
+    message("EpiModel 1.7.0 onward renamed the birth summary statistic b.flow to a.flow. See documentation for details.")
+  }
+
```
smjenness: Here and below for `param.icm`: `b.flow` is not a parameter passed into these parameter functions; it is a summary stat that gets output from the model. So these lines (143 to 146) should be removed, same for below.

PR #296, opened 2018-11-14
Changed default plotting behavior for dyanmic=FALSE network diagnostic plotting

andsv2:
Lowess smoothing is default plotting with option of overlaying individual simulation results over this. No updated unit test as I'm not sure how this change can error out.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -935,6 +935,13 @@ draw_means <- function(x, y, mean.smooth, mean.lwd,
 #' # Only formation diagnostics are available to plot
 #' plot(dx1, stats = "edges")
 #' plot(dx1, stats = c("edges", "concurrent"))
+#' plot(dx1, stats = c("edges", "concurrent"), sim.lines = FALSE)
+#' plot(dx1, stats = c("edges", "concurrent"), sim.lines = TRUE)
+#' plot(dx1, stats = c("edges", "concurrent"), sim.lines = FALSE, plots.joined=FALSE)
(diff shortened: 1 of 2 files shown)


R/plot.R
```diff
@@ -1135,14 +1135,12 @@ plot.netdx <- function(x, type = "formation", method = "l", sims, stats,
...
-          sim.lines <- FALSE
-        } else {
-          sim.lines <- TRUE
-        }
+      if(missing(sim.lines)){
+        sim.lines <- FALSE
       }
+       else {
+          simlines <- TRUE
+       }
```
smjenness: Note the variable name error

smjenness:
Additionally, this code block yields a warning message, probably related to trying to calculate quantiles. Please double check and fix as necessary.

```r
nw <- network.initialize(100, directed = FALSE)
formation <- ~edges
target.stats <- 50
coef.diss <- dissolution_coefs(dissolution = ~offset(edges), duration = 25)

# Estimate the model
est <- netest(nw, formation, target.stats, coef.diss, verbose = FALSE)

dx1 <- netdx(est, nsims = 1e4, dynamic = FALSE,
             nwstats.formula = ~edges + meandeg + concurrent)
plot(dx1, stats = "edges")
```

R/plot.R
andsv2: I'm not sure if this was the best solution, but I put dynamic=TRUE conditions when quantile calculations are being done; would it make more sense/cleaner to simply separate the two conditions from each other (dynamic = TRUE vs dynamic = FALSE).

Issue #297, opened 2018-11-15
Netsim Plots Erroring with Unspecified Ylim Message


connor-vanmeter:
Example of plot that led to "unspecified ylim" error message:
```
par(mar = c(3,3,1,1), mgp = c(2,1,0))
plot(sim, c("s.num","e.num","i.num","r.num"), type="epi",
     mean.col = 1:4, mean.lwd = 1, mean.smooth = FALSE,
     qnts = 1, qnts.col = 1:4, qnts.alpha = 0.25, qnts.smooth = FALSE,
     legend = TRUE)
```

PR #298, opened 2018-11-15
Update R/plot.R - Add na.rm option to draw_means helper function

connor-vanmeter:
The na.rm=TRUE option was not specified in the draw_means helper function when getting the min and max mean value for specified plot terms. As a result, an error message was received explaining that there was an unspecified ylim when plotting. NA values were being returned as the mean min and max before being set as the ylim lower and upper bound. This is not an issue in the draw_qnts as NA values are dealt with using na.rm in the main body of the function, and was only an issue in draw_means when nsims=1.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -851,8 +851,8 @@ draw_means <- function(x, y, mean.smooth, mean.lwd,
           col = mean.pal[j], lty = mean.lty[j])
     }
     else {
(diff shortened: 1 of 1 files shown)


connor-vanmeter:
Addresses Issue #297

connor-vanmeter:
Closes #297

Issue #325, opened 2019-11-12
plot.netsim default color specification


smjenness:
@connor-vanmeter would you please work on this in the next couple weeks?

There is a problem that the default colors for the mean lines and quantile bands do not show up when we explicitly specify the plot output with `y` when that output length is greater than 3 variables. The plot palettes should draw the the longer vector of palette colors in cases where the user wants to plot 4 or more variables in a single plot.

Case below:

```r
nw <- network.initialize(n = 100, directed = FALSE)
formation <- ~edges
target.stats <- 50
coef.diss <- dissolution_coefs(dissolution = ~offset(edges), duration = 20)
est1 <- netest(nw, formation, target.stats, coef.diss, verbose = FALSE)

# Epidemic model
param <- param.net(inf.prob = 0.3)
init <- init.net(i.num = 10)
control <- control.net(type = "SI", nsteps = 100, nsims = 5, verbose.int = 0)
mod1 <- netsim(est1, param, init, control)

# Four variables are able to be plotted, but the final si.flow does not display
mod1
plot(mod1, y = c("s.num", "i.num", "num", "si.flow"), legend = TRUE)

# Here we explicitly specify the plot colors, and all four variables display
plot(mod1, y = c("s.num", "i.num", "num", "si.flow"), mean.col = 1:4, qnts.col = 1:4, legend = TRUE)
```

connor-vanmeter:
Sure Sam, will get started on it

smjenness:
Closed with #327

PR #327, opened 2019-11-15
Update plot.R - plot.netsim color palette

connor-vanmeter:
The plot.netsim function originally displayed three specified plot variables when mean.col and qnts.col were not user-specified due to hardcoding of the color palette for three colors.
Combined "Set1", "Set2", and "Set3" color brewer palettes for the fix to allow the max number of plotted variables to be 29 when user does not specify mean.col or qnts.col.

Resolves Issue #325

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -2214,8 +2214,11 @@ plot.netsim <- function(x, type = "epi", y, popfrac = FALSE, sim.lines = FALSE, 
     ## Color palettes ##
 
     # Main color palette
-    bpal <- brewer.pal(3, "Set1")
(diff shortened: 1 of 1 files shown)


andsv2:
This all works on my side locally. I don't think I have the ability to merge though.

My only comment would be, is there anyway to base the number of colors pulled to be based on the number of stats requested? Instead of hardcoding a pool of around 30 colors I mean (though I agree that no one is likely to go beyond that).

smjenness:
I think Conor's approach of combining the three palettes should be fine, even if there are some colors that we do not end up using in the combined palette.

PR #350, opened 2020-03-19
Proposition: Harmonize netsim and allow raw outputs

AdrienLeGuillou:
Create a `process_out.net` function taking a list of `dat` object and
outputing a formatted `netsim` object. This allows harmonizing how the
mono and multicore version of `netsim` produce output.

In addition, a `raw_output` parameter is added to `netsim` (with default
== FALSE). If set to `TRUE`, a list of `dat` object is outputted instead
of a formatted `netsim` object.

Explanation:

The first part of the commit is a simplification of the code. The second part makes it easier to debug `netsim` as you can study the internal data. Furthermore, this could simplify greatly some checkpointing capabilities present in EpiModelHPC that for now require to deconstruct the processed output.

This is purely a proposition and comments are welcome


smjenness:
@AdrienLeGuillou please update from latest branch merge and address merge conflicts. Also please provide a minimal example demonstrating this new functionality.

AdrienLeGuillou:
exemple:

``` r
devtools::load_all()

nw <- network.initialize(n = 100, directed = FALSE)
nw <- set.vertex.attribute(nw, "race", rbinom(50, 1, 0.5))
est <- netest(nw, formation = ~edges + nodematch("race"), target.stats = c(25, 10),
              coef.diss = dissolution_coefs(~offset(edges), 10, 0),
              verbose = FALSE)


param <- param.net(inf.prob = 0.3, act.rate = 0.5)
init <- init.net(i.num = 10)
control <- control.net(type = "SI", nsims = 6, ncores = 2,
                       nsteps = 100, verbose = FALSE)

# Using the same syntax as before: same output as before
mod_orig <- netsim(est, param, init, control)

# Adding the new `raw_output` parameter
mod_raw <- netsim(est, param, init, control, raw_output = TRUE)

# mod_raw is a list of `dat` object
str(mod_raw[[1]], max.level = 1)
## List of 10
##  $ param    :List of 4
##   ..- attr(*, "class")= chr [1:2] "param.net" "list"
##  $ init     :List of 1
##   ..- attr(*, "class")= chr [1:2] "init.net" "list"
##  $ control  :List of 36
##   ..- attr(*, "class")= chr [1:2] "control.net" "list"
##  $ attr     :List of 6
##  $ stats    :List of 2
##  $ temp     :List of 2
##  $ nw       :List of 5
##   ..- attr(*, "class")= chr [1:2] "networkDynamic" "network"
##   ..- attr(*, "formation")=Class 'formula'  language nw ~ edges + nodematch("race")
##   .. .. ..- attr(*, ".Environment")=<environment: 0x55fb6b0b2130> 
##   ..- attr(*, "dissolution")=Class 'formula'  language nw ~ offset(edges)
##   .. .. ..- attr(*, ".Environment")=<environment: 0x55fb6b0ab090> 
##   ..- attr(*, "monitor")=Class 'formula'  language nw ~ edges + nodematch("race")
##   .. .. ..- attr(*, ".Environment")=<environment: 0x55fb6b0a5e20> 
##   ..- attr(*, "stats")= 'mcmc' num [1:100, 1:2] 46 46 44 43 42 39 39 43 39 37 ...
##   .. ..- attr(*, "dimnames")=List of 2
##   .. ..- attr(*, "mcpar")= num [1:3] 1 100 1
##   ..- attr(*, "coef.form")= Named num [1:2] -7.307 -0.387
##   .. ..- attr(*, "names")= chr [1:2] "edges" "nodematch.race"
##   ..- attr(*, "coef.diss")= num 2.2
##   ..- attr(*, "constraints")=Class 'formula'  language ~.
##   .. .. ..- attr(*, ".Environment")=<environment: 0x55fb6b09f100> 
##   ..- attr(*, "changes")= num [1:583, 1:4] 1 1 1 1 1 1 1 1 1 2 ...
##   .. ..- attr(*, "dimnames")=List of 2
##  $ nwparam  :List of 1
##  $ epi      :List of 4
##  $ nw.update:List of 1

# The processing of the list can be done outside the function
mod_processed <- process_out.net(mod_raw)

# Also works on a subset of the simulation
mod_sub <- process_out.net(mod_raw[2:4])
```

AdrienLeGuillou:
already with only this change, it becomes super easy to restart a simulation from a `dat` object and I would like to implement other convenience functions to make it even simpler. This way EpiModel could handle stoping and restarting simulations without the help of EpiModelHPC and it would provide more flexibility as the user could inspect the `dat` objects any time he wants.

``` r
dat <- mod_raw[[1]]

dat$control$initialize.FUN <- function(dat, ...) return(dat)
dat$control$start <- dat$control$nsteps + 1
dat$control$nsteps <- dat$control$nsteps + 100
EpiModel:::netsim_loop(dat, dat$param, dat$init, dat$control, 1)
```

smjenness:
Will this avoid the need to have a separate function for the initialization module when we want to "reinitialiize". 

For example, in core EpiModel, do we need changes here:
https://github.com/statnet/EpiModel/blob/master/R/net.mod.init.R#L80-L91

And in EpiModelHIV, do we need changes here:
https://github.com/statnet/EpiModelHIV/blob/CombPrev/R/mod.initialize.R#L303-L342

Also, I think I would prefer the raw_output setting to be embedded within the control settings. If you think there is a case for it as a top-level argument to netsim, tell me why.

AdrienLeGuillou:
My idea would be in the end to have a way to restart without anything but `dat` and `at`. So yes `reinit` would not be necessary anymore.

With the code here, nothing needs to change. I made sure that if we run `netsim` as before, the outputs stay the same and nothing needs to be changed. 

Overall, I think it would be good to have some low level flexible functions to easily tackle problems such as checkpointing, parallelization in a variety of different environments and anything an advanced user might want to do. But in the meantime, I believe it is important to keep the  "frontend" functions like `netsim` as similar as possible to what we have now in terms of inputs and outputs.  

No problem with it being in the `control` object, making the changes now.

martinamorris:
This sounds great.

smjenness:
@AdrienLeGuillou It looks like you removed raw_output from the netsim arguments, but then did not add it to the control.net arguments. Let's explicitly list it there with a default of FALSE.

Also, can you remove this reinit functionality from the init mod, since it's no longer needed?

Finally, I feel like process_out.net should only accept a full netsim class object (or maybe we create a class, netsim_raw for pre-processed output, and then the class changes to netsim after processing) and then it should have an argument, sims, to specify which subset of simulations to process. What do you think?

AdrienLeGuillou:
Argument added to control.net.

For reinit functionality, I don't think that these modifications alone are enough to remove the current functionalities without breaking the current usage. 

Not sure I understand what you mean. I feel that if process_out.net accepts only `netsim` objects, then it does the same thing as `merge.out` and requires the `dat` object to be modified before being passed to it. But the main idea is to have a way to get the unmodified `dat` at the end without disturbing the current `netsim` functionalities.

I think this PR should not be included in the next release of EpiModel. But if the idea in general to have more low-level functions to address the issues mentioned above are of interest to you, I can make a draft of the full set of additions I envision and what problem/limitation they would solve so we can discuss that together before I implement a more comprehensive patch. What do you think?

smjenness:
Ok, I'll merge this PR now and we can discuss broader API changes related to this next week! Thanks!

Issue #341, opened 2020-03-10
Network PIDS: Male, Female


andsv2:
In `init_pids` does it make sense to change

https://github.com/statnet/EpiModel/blob/807bd5a7bdf88df7f194e48d1f314638cc0192b5/R/net.mod.init.R#L260

to "Grp1, Grp2" or something similar.

andsv2:
Answer: no. However, how does `init_pids` work with `networkLite` object from `tergmLite`.

andsv2:
Answer: It doesn't.

martinamorris:
just my 2cents on this already closed issue.  there are other possible applications for diffucion on bipartite graphs (think kids and schools).  so for the purposes of generalization, it might be worth considering a generic bipartite labeling (b1 and b2 is what `ergm` uses).

smjenness:
Yes, let's keep this open to revisit it once the tergmLite integration is complete.

smjenness:
Closed in #354

Issue #307, opened 2018-12-17
Persistent IDs should be consistent and integers across model parameterizations


smjenness:
here's the current storage for bipartite networks:

```
head(get_transmat(sim1, sim = 1), 10)
   at         sus         inf infDur transProb actRate finalProb
1   2 5ff4952cb01 5ff373df9b6     27       0.1       5   0.40951
2   2 5ff1dfbdb3c 5ff5278d95b     28       0.1       5   0.40951
3   2 5ff5831c407 5ff7f48d4b9      7       0.1       5   0.40951
4   2 5ff5f80368a 5ff3153849a     24       0.1       5   0.40951
5   2 5ff6fc977ef 5ff3153849a     24       0.1       5   0.40951
6   2  5ffcda6ab1 5ff4a407a09     40       0.1       5   0.40951
7   3 5ff5831c407 5ff7f48d4b9      8       0.1       5   0.40951
8   3 5ff6cf91e6c   5ff681a44     50       0.1       5   0.40951
9   4 5ff137eb1db 5ff4952cb01      2       0.1       5   0.40951
10  4 5ff12fe42b4 5ff110cca80     10       0.1       5   0.40951
```

Just use integers instead, similar to EpiModelHIV's `uid`.

skyebend:
If these were generated using the network dynamic persistant id functionality, i think you can instead provide ids, or tell it to use a pre-existing vertex attribute (even vertex.names, if they are already unique integers)

smjenness:
Closed in #354

Issue #333, opened 2020-02-17
Task 4: Create a tergmLite version of the network_update module function


andsv2:
### Goals
 -  tergmLite is used for network simulation, this version of the module would replace the tergm/networkDynamic version for more efficient simulation 
 - Main speed improvements with edgelist-only version of add and delete vertices 

### TODO
- [x] Port functionality from EpiModelHIV 
- [x] Address all remaining errors in R CMD Check
- [x] Consolidate init_status functions
- [x] Check/edit init module
- [x] Check/edit infection module
- [x] Check/edit resim_nets
- [x] Check/edit nwupdate.net
- [x] Rename prevalence module
- [x] Rename nw.update.net module/fx
- [x] Rename 2 group functions to X.2g.net
- [x] Change storage location for `nwupdate.net` from `dat$nw.update...` to `dat$temp...`; and decide which of these are necessary (versus on-the-fly queries)
- [x] Rename `depend` control setting. `resimulate.network`?
- [x] Figure out what to do with persistent IDs (#307, #341); also, PIDs are not getting updated (at least for 2g models)
- [x] Start on unit testing
- [x] Move new nodal attr updates to `dat$attr` from `nwupdates.net` to `arrivals.net`
- [x] Change variable name `bip`, with T/F value, in crosscheck to nGroups, with 1/2 value
- [x] Simplify copying attr back/forth from dat$attr to nw: instead of limiting it to only terms on the ergm formula, do it for all attributes originally passed on the network
- [x] Test/adapt serosorting (statOnNw) model for tergmLite
- [x] Remove duplicate setting of dat$attr code in arrivals.net; we should not need to look up based on use of tergmLite
- [x] Address hard-coding of group numbers to 2, like this `length(unique(get.vertex.attribute(nw, "group")))`
- [x] Address cases where group is on the nw as an attribute but not in the ergm formula; and also address cases where other attributes (race, age) are passed in on the network but not on the ergm formula. 
- [x] `tergmLite` replaces save.network and save.transmat. If FALSE, then save these two things. If TRUE, don't save them.
- [x] Check on why the calls to `test_net` in test-net-long unit test script are failing on R devel
- [x] merge.netsim doc issue (#342); although if save.network is removed, the fix is different
- [x] Add unit tests with attribute copying based on examples at the end of the EpiModel test script

smjenness:
@andsv2 Running your test scripts, this still does not work with netsim:

```r

num <- 200
[...]

PR #354, opened 2020-03-22
Update handling of PIDs

smjenness:
Addresses this issue from #333 
> Figure out what to do with persistent IDs (#307, #341); also, PIDs are not getting updated (at least for 2g models)


smjenness:
Closes #307 
Closes #341

PR #356, opened 2020-03-22
EpiModel Overhaul Updates for 3/22/20

smjenness:
Handling updates in #333


R/net.mod.nwupdate.R
```diff
@@ -75,7 +75,7 @@ nwupdate.net <- function(dat, at) {
     }
 
     ## Recovery
-    idsRecov <- dat$nw.update$rec$idsRecov
+    idsRecov <- dat$temp$idsRecov
     recovState <- ifelse(dat$control$type == "SIR", "r", "s")
```
smjenness: Can we figure out a way to do this without having to write out idsRecov? One solution, is to look up, in the nwupdate fx the setdiff of ids with a recov status on attr versus the network object.

smjenness:
@andsv2 Let's also change the name of the `tgl` control setting to `tergmLite`.

R/net.mod.nwupdate.R
andsv2: I was going to suggest a time recovered attr like for infection (time infected), but the above works too without needing to add another tracker.

smjenness:
@andsv2 Looks great. Once the query of new recoveries in nwupdate.net (instead of writing to dat$temp) is done, I think this will be ready to merge into _s4.

R/net.mod.nwupdate.R
```diff
@@ -66,9 +66,12 @@ nwupdate.net <- function(dat, at) {
     }
 
     ## Recovery
-    idsRecov <- dat$temp$idsRecov
-    recovState <- ifelse(dat$control$type == "SIR", "r", "s")
     status <- dat$attr$status
+    attr.status <- which(status == "r")
+    nw.status <- get.vertex.attribute(dat$nw, "status")
```
smjenness: This would work for SIR models but not SIS models, right?

andsv2:
All Travis checks passing and lingering issues (without Travis flags) addressed. Merging into _s4 branch.

PR #358, opened 2020-03-24
EpiModel Overhaul updates for 3/24/2020

andsv2:
Addresses the following issues in #333 :

- [x] Change variable name `bip`, with T/F value, in crosscheck to nGroups, with 1/2 value
- [x] Simplify copying attr back/forth from dat$attr to nw: instead of limiting it to only terms on the ergm formula, do it for all attributes originally passed on the network
- [x] Test/adapt serosorting (statOnNw) model for tergmLite
- [x] Address cases where group is on the nw as an attribute but not in the ergm formula; and also address cases where other attributes (race, age) are passed in on the network but not on the ergm formula.


R/net.utils.R
```diff
@@ -197,23 +197,75 @@ color_tea <- function(nd, old.var = "testatus", old.sus = "s", old.inf = "i",
...
 #'
 copy_toall_attr <- function(dat, at, nwterms) {
-  otha <- names(dat$nw$val[[1]])
-  otha <- otha[which(otha %in% nwterms)]
-  if (length(otha) > 0) {
-    for (i in seq_along(otha)) {
-      va <- get.vertex.attribute(dat$nw, otha[i])
-      dat$attr[[otha[i]]] <- va
-      if (at == 1) {
+
```
smjenness: what role does the nwterms arg play, given that we want to copy all attr from nw to attr?

andsv2: In the CRAN version of ```get_attr_prop```, there is an input parameter ```only.formula``` to limit updating the attribute list to this attributes in the formation model (```fterms```). This would distinguish between the two. Can change this to all attributes as the only case and remove that language.

smjenness: Just to be clear, if there is an open population model, and an attribute passed on the network that is not in the ERGM formula, these attributes should be added/updated following the same rules as attributes on the ERGM formula.

andsv2:
@smjenness 

```get_attr_prop```: Updated function so that if there are no network attributes, get_attr_prop will not run. Also changed attribute names that are tracked to be specific to tergmLite: TRUE or FALSE.

```auto_update_attr```: Change function name to account for new functionality. Update proportion lookup based on tergmLite: TRUE or FALSE. Attribute partitioning now based on length of newNodes (tergmLite = FALSE) or total nodes (tergmLite = TRUE).

```copy_toall_attr```: Revert back to master. 

```nwupdate.net```: Reorg network bookkeeping: Arrivals now occurs before Departures in order to make sure attribute list is correct length.

```initialize.net```: Added check for when there are no attributes on the network.

Issue: test-net-long.R causes R to abort. Issue with formation model?

PR #366, opened 2020-04-08
Getters and setters functions for the `dat` object

AdrienLeGuillou:
Addition of `get_` and `set_` function for `dat$attr`, `dat$param`, and `dat$epi`.
Each of these function derives from generic `set_dat_elt` and `get_dat_elt`


AdrienLeGuillou:
I added the ability for `get_` functions to access the full list, an element of the list or a subset of the list. One use case I have for this is the ability to write modules this way:

``` r
new_module <- function(dat, at) {
  with(c(
    get_attr(dat, c("age", "status"), TRUE),
    get_param(dat, c("hiv.prob", "mod_inf"), TRUE),
    get_epi(dat, "s", TRUE)),
  {
    age <- age + 1
    status[!status] <- rbinom(length(status[!status]), 1, hiv.prob)
    s[[at]] <- sum(!status)

    dat <- set_epi(dat, "s", s)
    dat <- set_attr(dat, "status", status)

    return(dat)
  })
}
```
This approach has three major benefits. 

- First, it takes advantage of the vectorized way the `get_` functions can work and in case of repetitive access of the elements, the access time is reduced (thanks to with). 
- Second, it makes obvious at the top of the modules what elements are going to be used and at the end what is going to be changed in the internal object (I wished I had that when I explored EpiModelHIV the first time).
- Finally, inside the modules the names are completely devoid of $ and are the same of their internal counterparts.

smjenness:
Thanks Adrien. Excellent work. Some comments:

* get_attr should have a v parameter, similar to get.vertex.attribute, which is a vector of IDs to to subset the extraction. The default is NULL, with no subsetting. Complain with an informative error if the attr is not on the list.
* set_attr: I am not certain if this functionality is possible, but if the attr does not exist on dat$attr at the time of setting, create it new. Complain with a relevant error if the new values passed in would violate logic (e.g., more values than the population size)
* get_epi: similarly, add an at argument to allow for subsetting the epi list by a vector of time steps, with the NULL default with no subsetting.
* set_epi: if the epi does not exist dat$epi at the time of setting, create it new. In the case where it is set at a subsequent time step (e.g., I expect this to be used within a recovery module that gets run for the first time when at == 2) initialize the vector with NAs of the length of the control$nsteps, then set the value for the current time step.
* simliar get functions as get_param for init and control outputs.

AdrienLeGuillou:
I think I can implement all of this without problems. I have some questions: 
- do you prefer having the creation of new elements in the same function (`set_`) or in a different one (`add_attr` for example)? One function is more flexible but two functions ensure a clearer code as you can always distinguish between updating and creating. Tell me which way you want me to go. 
- for the subsetting vector, should I make it a vector of ID's only or should I allow it to be a logical vector of the same size as the list we try to access? In order to have the same capabilities as a base R vector?

smjenness:
> * do you prefer having the creation of new elements in the same function (`set_`) or in a different one (`add_attr` for example)? One function is more flexible but two functions ensure a clearer code as you can always distinguish between updating and creating. Tell me which way you want me to go.

I can see the potential lack of clarity with collapsing this into a single function, but I think I would like a single function, so users don't have to do this in modules:

```r
if (is.null(get_attr(dat, "age")) {
add_attr(dat, "age, ...)
} else {
set_attr(dat, "age", ...)
}
```

Ok?

smjenness:
> * for the subsetting vector, should I make it a vector of ID's only or should I allow it to be a logical vector of the same size as the list we try to access? In order to have the same capabilities as a base R vector?

I can see the utility of supporting either. Good idea!

AdrienLeGuillou:
> > * do you prefer having the creation of new elements in the same function (`set_`) or in a different one (`add_attr` for example)? One function is more flexible but two functions ensure a clearer code as you can always distinguish between updating and creating. Tell me which way you want me to go.
> 
> I can see the potential lack of clarity with collapsing this into a single function, but I think I would like a single function, so users don't have to do this in modules:
> 
> ```r
> if (is.null(get_attr(dat, "age")) {
> add_attr(dat, "age, ...)
> } else {
> set_attr(dat, "age", ...)
> }
> ```
> 
> Ok?

I think the question is how do you want to present the tutorial for modules. 
If we go with the two functions, the idea would be to "declare" all the elements at once  with `add_`, for example in the initialization module  and the `set_` would be used in the other modules. This way if you want to know the full amount of information in `dat` the init module will be enough. 

The unique function makes it closer to what we have now. One risk is a user making a typo in the name of one attribute, resulting in the creation of a new one instead of updating.

``` r
# if dat$attr contains only "age"
# one function situation
dat <- set_attr(dat, "aje", value) # silently create an "aje" element

# two function situation 
dat <- set_attr(dat, "aje", value) # informative error 
```

In any case, the two functions will exist in order to have the initialization optimization ready. The question is whether to have `set_` as a wrapper function or not.

smjenness:
> In any case, the two functions will exist in order to have the initialization optimization ready. The question is whether to have `set_` as a wrapper function or not.

The initialization module is one of those that confuses most users, so I would like to have the option to directly initialize these attributes in the relevant modules within the time loop, e.g. at time step 2. Again, I know it's not the best coding practice here, but that's the tradeoff we're dealing with

AdrienLeGuillou:
the initialization module was just an example. 
In a module it would:

``` r 
# two functions
if (at == 2) {
  dat <- add_attr(dat, "age") # initialize empty "age" 
}

dat <-  set_attr(dat, "age", new_values) # update "age"

# one function
dat <-  set_attr(dat, "age", new_values) # check if "age" exist each time, create if not, then update 
```

smjenness:
Let's do the second option, one function for both, for now.

AdrienLeGuillou:
Here is a modified version of this implementation including the requested changes.

I also added a `set_epi_at` function that takes a timestep (unique) to allow the equivalent of:

``` r
dat$epi$x[[at]] <- 4
```
with 

``` r
dat <- set_epi_at(dat, "x", at, 4)
```

this function adds the required numbers of NA at the end of `x` if it is not long enough (for example after a restart where `dat$control$nsteps` would have changed

AdrienLeGuillou:
I am working on some tests before a review is required

R/net.dat.R
```diff
@@ -0,0 +1,385 @@
...
+
+  return(out)
+}
+
+#' @rdname dat_get_set
+#' @export
+add_epi <- function(dat, item) {
+  if (item %in% names(dat[["epi"]]))
+    stop(paste("Cannot create the Epidemic outpout, ", item,
+               ": exists already"))
```
smjenness, line 163: typo

R/net.dat.R
```diff
@@ -0,0 +1,385 @@
...
+  return(dat)
+}
+
+#' @rdname dat_get_set
+#' @export
+set_epi <- function(dat, item, value) {
+  if (!item %in% names(dat[["epi"]])) {
+    message(paste("Creating Epidemic output: '", item,
+               "'in the Epidemic output list of the Master list object (dat)"))
+
```
smjenness, line 176: I don't want these messages. They will just clutter the console if there are 50 of them. Please remove

R/net.dat.R
```diff
@@ -0,0 +1,385 @@
...
+               "' in the attributes list of the Master list object (dat)"))
+
+
+    dat <- add_attr(dat, item)
+  }
+
+  if (length(value) != length(dat$attr$active))
+    stop(paste0(
+      "When trying to edit the ", `item`, " nodale attribute: The size",
+       " of the `value` vector is not equal to the number of node in
```
smjenness, line 101: another typo: nodale

R/net.dat.R
```diff
@@ -0,0 +1,385 @@
...
+#' @export
+get_epi_list <- function(dat, item = NULL) {
+  if (is.null(item)) {
+    out <- dat[["epi"]]
+
+  } else {
+    missing_item <- setdiff(item,names(dat[["epi"]]))
+    if (length(missing_item) > 0)
+      stop(paste("There is no Epidemic output called",
+                 paste(missing_item, collapse = ", "),
```
smjenness, line 120: maybe let's call it epi output instead of Epidemic output

R/net.dat.R
```diff
@@ -0,0 +1,385 @@
...
+  return(dat)
+}
+
+#' @param at timestep where to add the new value for the Epidemic Outuput `item`
+#' @rdname dat_get_set
+#' @export
+set_epi_at <- function(dat, item, at,  value) {
+  if (length(at) != 1 || !is.numeric(at))
+    stop("`at` must be numeric and of length one")
+
```
smjenness, line 191: prefer that you wrap this in brackets, just for stylistic consistency

R/net.dat.R
```diff
@@ -0,0 +1,385 @@
...
+#' @export
+get_param_list <- function(dat, item = NULL) {
+  if (is.null(item)) {
+    out <- dat[["param"]]
+
+  } else {
+    missing_item <- setdiff(item,names(dat[["param"]]))
+    if (length(missing_item) > 0)
+      stop(paste("There is no parameters called",
+                 paste(missing_item, collapse = ", "),
```
smjenness, line 222: There are no parameters named

R/net.dat.R
```diff
@@ -0,0 +1,385 @@
...
+
+  out <- dat[["param"]][[item]]
+
+  return(out)
+}
+
+#' @rdname dat_get_set
+#' @export
+add_param <- function(dat, item) {
+  if (item %in% names(dat[["param"]]))
```
smjenness, line 246: I'm not sure if we would ever want to add/set parameters within modules. These are always just pulled from the param list. Can you think of a reason?

AdrienLeGuillou: I supposed you wanted to use this kind of syntax for all interaction with the `dat` object (i.e. also in the `param` functions. Can remove otherwise.

R/net.dat.R
AdrienLeGuillou: what do you mean? If I do:

``` r
names(param.net())
[1] "act.rate" "vital"   

```

R/net.dat.R
AdrienLeGuillou: corrected

R/net.dat.R
AdrienLeGuillou: removed

R/net.dat.R
AdrienLeGuillou: corrected

R/net.dat.R
AdrienLeGuillou: done

R/net.dat.R
AdrienLeGuillou: done

R/net.dat.R
smjenness: Grammar edit: "There are no parameters named" instead of "There is no parameters called"

R/net.dat.R
smjenness: No problem leaving them in there, for consistency at the least. Maybe some users will have a need.

Issue #352, opened 2020-03-20
Fix print.netsim output


smjenness:
The print output for netsim objects needs some updating.

For example, this model:
```r
nw <- network.initialize(n = 100, directed = FALSE)
nw <- set.vertex.attribute(nw, "race", rbinom(50, 1, 0.5))
est <- netest(nw, formation = ~edges + nodematch("race"), target.stats = c(25, 10),
              coef.diss = dissolution_coefs(~offset(edges), 10, 0),
              verbose = FALSE)


param <- param.net(inf.prob = 0.3, act.rate = 0.5)
init <- init.net(i.num = 10)
control <- control.net(type = "SI", nsims = 6, ncores = 2,
                       nsteps = 100, verbose = FALSE)

# Using the same syntax as before: same output as before
mod_orig <- netsim(est, param, init, control)
mod_orig
```

generates this output:
```r
EpiModel Simulation
=======================
Model class: netsim

Simulation Summary
-----------------------
Model type: SI
No. simulations: 6
No. time steps: 100
No. NW groups: 1

Model Parameters
-----------------------
inf.prob = 0.3
act.rate = 0.5


Model Functions
-----------------------
arrivals.net = arrivals.FUN 
departures.net = departures.FUN 
infection.net = infection.FUN 
prevalence.net = prevalence.FUN 


Model Output
-----------------------
Variables: s.num i.num num si.flow
Networks: sim1 ... sim6
Transmissions: sim1 ... sim6
```

Fixes needed:
- [ ] reverse order of Model Function listings (it should be arrivals.FUN = arrivals.net)
- [ ] reduce double spacing around Model Functions block

smjenness:
Fixed in #370

Issue #353, opened 2020-03-20
Fix plot.netsim color coding for 2-group SIR models


smjenness:
MRE:

```r
num1 <- num2 <- 500
nw <- network.initialize(num1 + num2, directed = FALSE)
nw <- set.vertex.attribute(nw, "group", rep(1:2, each = num1))
formation <- ~ edges + nodematch("group")
target.stats <- c(400, 0)
coef.diss <- dissolution_coefs(dissolution = ~offset(edges), duration = 20)
est <- netest(nw, formation, target.stats, coef.diss)

# Parameters
init <- init.net(i.num = 10, i.num.g2 = 10, r.num = 5, r.num.g2 = 5)
param <- param.net(inf.prob = 0.4, inf.prob.g2 = 0.2,
                   rec.rate = 0.02, rec.rate.g2 = 0.02,
                   act.rate = 1)
control <- control.net(type = "SIR", nsteps = 250, nsims = 5, ncores = 5, tgl = TRUE)

sim <- netsim(est, param, init, control)
plot(sim)
```

<img width="899" alt="Screen Shot 2020-03-20 at 5 00 09 PM" src="https://user-images.githubusercontent.com/2521570/77205927-48794080-6acc-11ea-8248-df60b4c9daaa.png">

smjenness:
Fixed in #370

PR #370, opened 2020-04-09
printplot: Update to multiple group plotting and format of print.

andsv2:
Update to print and plot function. ```print.netsim```: requested formatting changes. ```plot```: color assignment based on number of epi stats plotted; colors now assigned by class of stat instead of unique colors for each stat.


smjenness:
Fixes #352 and #353

smjenness:
@andsv2 Perfect, thanks. Merging in

PR #373, opened 2020-04-10
Merge _s4 into _Overhaul

smjenness:


smjenness:
Error on Travis with one netdx unit test, does not reproduce locally, and only on the PR check on Travis. Merging anyway, and reevaluating from there.

Issue #320, opened 2019-05-17
Check print.control.net output for built-in 2-group models


smjenness:
For example: 

```r
num1 <- num2 <- 250
nw <- network::network.initialize(num1+num2, directed = FALSE)
nw <- network::set.vertex.attribute(nw, "group", rep(c(1,2), each = num))
formation <- ~ edges + nodematch("group") + nodefactor("group")
target.stats <- c(150, 125, 62.5)
coef.diss <- dissolution_coefs(dissolution = ~ offset(edges), duration = 50)
est2 <- netest(nw, formation, target.stats, coef.diss)
dx2 <- netdx(est2, nsims = 5, nsteps = 250)
print(dx2)

#Parameters
init <- init.net(i.num = 5, i.num.g2 = 5)
param <- param.net(inf.prob = 0.1, inf.prob.g2 = 0.2,
                   rec.rate = 0.02, rec.rate.g2 = 0.02,
                   act.rate = 5)
control <- control.net(type = "SIS", nsteps = 100, nsims = 5)
control
```

provides this output:
```r
Network Model Control Settings
===============================
type = SIS
nsteps = 100
start = 1
nsims = 5
ncores = 1
rec.rand = TRUE
a.rand = TRUE
d.rand = TRUE
tea.status = TRUE
use.pids = TRUE
module.order =
save.nwstats = TRUE
nwstats.formula = formation
delete.nodes = FALSE
save.transmat = TRUE
save.network = TRUE
verbose = TRUE
verbose.int = 1
skip.check = FALSE
attr.rules = <list>
f.names = arrivals.FUN departures.FUN infection.FUN recovery.FUN get_prev.FUN
f.args = arrivals.net departures.net infection.net recovery.net get_prev.net
```

shouldn't this list the names of the 2-group functions under f.args?

smjenness:
@andsv2 Reminder on this

smjenness:
Actually, revert the last commit that updates crosscheck.net. I wasn't considering that print.control.net gets run before crosscheck.net. Instead, I would like print.control.net just to print out the module names in the order in which they are evaluated, and not print the function names.

So it should look like this, but with updates on the module names (removing the initialize and verbose ones) and ordering based on the Overhaul work:
```r
Network Model Control Settings
===============================
type = SIS
nsteps = 100
start = 1
nsims = 5
ncores = 1
rec.rand = TRUE
a.rand = TRUE
d.rand = TRUE
tea.status = TRUE
[...]

PR #377, opened 2020-04-14
EpiModel: Update to print for two-group modules

andsv2:
Update so that print correctly prints the two-group functions when two group model is specified.

https://github.com/statnet/EpiModel/issues/320


R/net.inputs.R
```diff
@@ -781,6 +781,28 @@ crosscheck.net <- function(x, param, init, control) {
       }
     }
 
+    ## Update f.names/args
+    if (flag == 0) {
+    if (control$type != "SIR" && !is.null(control$type)) {
```
smjenness, line 786: Fix indentation

smjenness: Nix that. See updated comments on #320

R/print.r
```diff
@@ -456,8 +456,9 @@ print.control.net <- function(x, ...) {
   if(is.null(x$module.order)) {
     cat("Dynamic Modules:", funToPrint)
   } else {
-    cat("Dynamic Modules:", intersect(x$module.order,
-                                      funToPrint))
+    order <- unlist(lapply(funToPrint, function(y) which(y == x$module.order)))
+    funToPrint <- funToPrint[order]
+    cat("Dynamic Modules:", funToPrint)
```
andsv2, line 461: @smjenness can I ask what the difference between this and printing module.order should be (identical at this point). Should it just print the module order, or the module order then concatenated with remaining modules that run? Hopefully this makes sense.

smjenness: the print function should print out all the modules (except for initialize and verbose) in the order in which they are run within the netsim time loop.

andsv2: @smjenness updated the function but I'm not sure if this situation would ever exist: print takes module order and then concatenates with the default order of the other modules to get the full module run order; this is in case someone doesn't pass full list or modules into module.order parameter. If this is over-gimmicked can revert to just looking at module.order.

smjenness: Thanks. If ready to review pls click that button on the PR

andsv2:
@smjenness ready for review.

PR #379, opened 2020-04-14
add append_attr and its tests

AdrienLeGuillou:


R/net.dat.R
```diff
@@ -162,6 +167,28 @@ set_attr <- function(dat, item, value) {
   return(dat)
 }
 
+#' @param n_new the number of new elements to append at the end of `item`
+#' @rdname dat_get_set
+#' @export
+append_attr <- function(dat, item, value, n_new) {
+  if (!item %in% names(dat[["attr"]])) {
```
smjenness, line 174: also, I prefer to keep function arguments with period case, so n.new instead of n_new

R/net.dat.R
```diff
@@ -162,6 +167,28 @@ set_attr <- function(dat, item, value) {
...
+
+  if (length(value) != 1) {
+    stop("`value` must be of length one.")
+  }
+
+  if (!is.numeric(n_new) || n_new < 0) {
+    stop("`n_new` must be numeric and greater than or equal to zero.")
+  }
+
+  dat[["attr"]][[item]] <- c(dat[["attr"]][[item]], rep(value, n_new))
```
smjenness, line 187: what about the case where we want to assign a vector. Say we have an attribute "group", with n.new= 5, and values of 11212. Can we allow for that possibility? So Value is either a single number that gets repeated n.new times or a vector of length n.new.

AdrienLeGuillou: I can make it work

R/net.dat.R
AdrienLeGuillou: ok

Issue #397, opened 2020-04-24
Wrapper function for set_transmat


smjenness:
Simplifies, wraps this code in the infection module

```r
    if (at == 2) {
      dat$stats$transmat <- del
    } else {
      dat$stats$transmat <- rbind(dat$stats$transmat, del)
    }
```

smjenness:
Closed with #398

PR #398, opened 2020-04-27
Addition of set_transmat internal function

andsv2:
Wrap appending of transmission matrix into a helper function, cleaning up infection.FUN modules.

Addresses #397


andsv2:
@smjenness for review.

Issue #392, opened 2020-04-21
Update delete_vertices to use dat as first arg


smjenness:
so delete_vertices(dat...) rather than delete_vertices(dat$el[1]], ...)

smjenness:
Closed, keeping functionality as-is.

PR #400, opened 2020-04-27
Update Delete Vertices Function

andsv2:
Update to delete_vertices. Depends on tergmLite. #392.


smjenness:
@andsv2 Closing because it makes sense to keep functionality as-is in tergmLite. Sorry about that.

PR #402, opened 2020-04-29
Add set_network function

andsv2:
Function behaves in the same way as set_ family of functions: operates on dat object and returns updated dat in place.


andsv2:
@smjenness currently trying to get R 4.0 set up on local machine; I believe this is the reason for package failure as it's passing locally. However wanted to get this PR open as I have a few Q's: 

1. Should this apply to tergmLite == T or F? Right now it is set up to handle both
2. If (1) holds, should there be an "override.tergmlite" input parameter for instances where this function is being used outside of EpiModel workflow (in examples is what I was thinking of; can't think of any actual use case).
3. Similar to the above, if this should have functionality for both tgl = T or F, in the case of tgl = T, the edgelist that is updated through delete_vertices, is this meant to return an object of class edgelist or matrix?

smjenness:
Closing this PR for now, as it is not a key priority for 2.0 and I think the revisions above got overly complicated. We'll keep the branch open and return to it once 2.0 is out.

Issue #361, opened 2020-03-29
Check overhaul against JSS article script


smjenness:
- [x] Download original JSS script and place into inst/ folder: https://www.jstatsoft.org/index.php/jss/article/downloadSuppFile/v084i08/v84i08.R
- [x] Run through original JSS script and make sure the deprecation messages are clear
- [x] Develop new JSS script `inst/EpiModelJSS2.R` that uses all the new API
- [x] Create `.onLoad` message for EpiModel 2.0 release
- [x] Create RMarkdown tutorial for EpiModel 1 –> 2 migration (will be posted on epimodel.org)

PR #403, opened 2020-05-01
Update to JSS

andsv2:
An update to the JSS script to account for EpiModel 2.0 workflow. Retains original JSS file for comparison of changes. Addresses #361.


R/net.inputs.R
```diff
@@ -658,6 +658,18 @@ crosscheck.net <- function(x, param, init, control) {
         }
       }
 
+      # Checks that .m2 syntax update to .g2
+      m2.param <- grep("m2", names(param), value = TRUE)
```
smjenness, line 662: Isn't this in reverse of what it should be? Shouldn't this stop check be catching if the user passes in `.g2` parameters?

R/net.inputs.R
```diff
@@ -658,6 +658,18 @@ crosscheck.net <- function(x, param, init, control) {
         }
       }
 
+      # Checks that .m2 syntax update to .g2
```
smjenness, line 661: Why is this done in crosscheck and not in param.net and init.net?

R/net.inputs.R
andsv2: I put that in there in case the user wasn't aware of the change as opposed to them neglecting to input the parameter. Can meld a helpful message into the original with something along the lines of "Note: mode language changed to group."

smjenness: ok, right, got it. resolving this.

R/net.inputs.R
andsv2: Checks moved to param and init respectively. Updated unit tests.

andsv2:
Afternoon @smjenness: before shooting this off for a full review some observations:
There are original JSS errors that I'm curious if including errors is wanted:
1. When type = NULL, does not inform user that all functions must be present in control.net, so netsim fails as some functions are missing.
2. When plotting old epi trackers, simply says can't find the epi name in tracker list.

As for teasing out expected errors, these are the ones that immediately come to mind, was curious if you had any others off the top of your head:
1. Inform user of updates to births/deaths.
2. Inform user of "type = NULL" in case of using user modules.
3. User is informed of of change from m2 to g2 language.
4. Depend -> Resimulate Network.

smjenness:
You have misinterpreted the first step in this issue. The original script should not be altered in any way.

smjenness:
I agree with those four types of error messages.

andsv2:
> You have misinterpreted the first step in this issue. The original script should not be altered in any way.

Is not the original JSS from the article the original in question. The JSS uploaded has been hard-reverted back to the original as it looks like there were some changes in regards to "births/deaths" -> "arrivals/departures" that weren't reverted back (mostly in comments).

smjenness:
I don't understand the statement above, perhaps because I am not familiar with the term "hard-revert". I'm not clear what reversion in any level means here. The original script needs to be download and placed in the folder, and tested against. Thanks!

andsv2:
> I don't understand the statement above, perhaps because I am not familiar with the term "hard-revert". I'm not clear what reversion in any level means here. The original script needs to be download and placed in the folder, and tested against. Thanks!

Apologies. By "hard-revert" I mean I downloaded the original script, copied it's contents and copied those into the JSS.R file in inst. That is the file that is now in this version and has been tested against. Any changes reflected in the commit are changes back to it's original state.

smjenness:
I see. Thanks for clarifying. Carry on...

smjenness:
Isn't `delete.nodes` gone now? This was not flagged when running the original JSS code:
https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS.R#L143

smjenness:
Similarly, why was this not removed from the JSS2 script? https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS2.R#L145-L146

smjenness:
It is unclear why you updated this first example in JSS2 to be a two-group model: https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS2.R#L33

The article shows it was demonstrating a one-group model, although stratified by this risk attribute.

smjenness:
Just curious what happens mechanistically when this structurally bipartite model from JSS1 is run: https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS.R#L94

smjenness:
This example does not run!? 

https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS2.R#L244-L254

smjenness:
Please check that the results produced by JSS2 are all qualitatively the same as results in the paper PDF.

smjenness:
Why does the JSS2 use old parameters like save.network? These were supposed to be updated for the JSS2 script. https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS2.R#L359-L363

smjenness:
This model also does not run...

https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS2.R#L352-L366

andsv2:
> Isn't `delete.nodes` gone now? This was not flagged when running the original JSS code:
> https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS.R#L143

Is it? I am looking through issues and not seeing it referenced. Is it a consequence of another change I am not seeing?

andsv2:
> It is unclear why you updated this first example in JSS2 to be a two-group model: https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS2.R#L33
> 
> The article shows it was demonstrating a one-group model, although stratified by this risk attribute.

Originally thought that "risk0" was original "mode1" assignment. Changed back to original example.

andsv2:
> Please check that the results produced by JSS2 are all qualitatively the same as results in the paper PDF.

All models updated to perform as they did in original paper.

andsv2:
> Similarly, why was this not removed from the JSS2 script? https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS2.R#L145-L146

I removed this in testing (used tergmLite = TRUE), however not sure why this would be removed as a matter of fact.

andsv2:
> Just curious what happens mechanistically when this structurally bipartite model from JSS1 is run: https://github.com/statnet/EpiModel/blob/JSS/inst/JSS/EpiModelJSS.R#L94. The parameter names clearly get updated but then what happens to the modes?

So looking at this after running it under full and lite tergm: total number of edges isn't stable just like in the paper, but in the opposite direction. However mean degree isn't stable at all either.

Issue #401, opened 2020-04-28
Use new tergmLite intialize/get/set_ functions in EpiModel workflow


smjenness:
With these new functions: https://github.com/statnet/tergmLite/pull/40

Replacements:
- [ ] `network_initialize` replaces `network.initialize`
- [ ] `set_vertex_attribute` replaces `set.vertex.attribute`
- [ ] `get_vertex_attribute` replaces `get.vertex.attribute`

andsv2:
Closed: https://github.com/statnet/EpiModel/pull/405

PR #405, opened 2020-05-05
Using new tergmLite functions

andsv2:
Update EpiModel workflow to use tergmLite function - network/get/set.

Addresses #401.

### DIFF
--- a/R/as.data.frame.R
+++ b/R/as.data.frame.R
@@ -157,7 +157,7 @@ as.data.frame.dcm <- function(x, row.names = NULL, optional = FALSE, run,
 #'
 #' \dontrun{
 #' ## Stochastic SI network model
-#' nw <- network.initialize(n = 100, directed = FALSE)
+#' nw <- network_initialize(n = 100, directed = FALSE)
 #' formation <- ~edges
 #' target.stats <- 50
 #' coef.diss <- dissolution_coefs(dissolution = ~offset(edges), duration = 20)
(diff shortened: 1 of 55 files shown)


R/as.data.frame.R
```diff
@@ -157,7 +157,7 @@ as.data.frame.dcm <- function(x, row.names = NULL, optional = FALSE, run,
 #'
 #' \dontrun{
 #' ## Stochastic SI network model
-#' nw <- network.initialize(n = 100, directed = FALSE)
+#' nw <- network_initialize(n = 100, directed = FALSE)
```
smjenness, line 160: remove all the `directed = FALSE` throughout since that is now the default with network_initialize

R/dendo.R
```diff
@@ -235,7 +235,7 @@ as.network.transmat <- function(x, ...){
   set.edge.attribute(net, "finalProb", tm$finalProb)
 
   net %v% "at" <- 0
-  set.vertex.attribute(net, "at", tm$at, v = tm$sus)
+  set_vertex_attribute(net, "at", tm$at, v = tm$sus)
```
smjenness, line 238: need an explicit object assignment here:
`net <- set_vertex_attribute(net, ...)`

R/as.data.frame.R
andsv2: Done. Also cleaned up formatting so all function calls are consistent.

smjenness: @andsv2 as noted, *please* mark these done with "resolve conversation"

PR #406, opened 2020-05-06
Remove groupids.

andsv2:
With get_ suite of functions, functionality with ```groupids``` has been taken over. Another function ```idgroup``` still exists however acts directly on a network object and is still used (though this function, currently, is only used once in entire EpiModel scripting).


smjenness:
@andsv2 : in the future, please resolve any merge requests before sending to me for review.

PR #415, opened 2020-05-15
Accessor Consistency

andsv2:
Updated "$" syntax in net.dat.R to brackets for consistency; was curious as well if we want to use accessor functions themselves.

### DIFF
--- a/R/net.dat.R
+++ b/R/net.dat.R
@@ -149,7 +149,7 @@ add_attr <- function(dat, item) {
                "': exists already"))
   }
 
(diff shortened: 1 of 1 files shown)


smjenness:
@andsv2 Yes, probably best to be consistent throughout and use the double brackets.
@AdrienLeGuillou thoughts?

smjenness:
For now though, perhaps best to punt this to post 2.0 so we can wrap up the overhaul.

AdrienLeGuillou:
As a rule I tend to prefer double bracket when not in interactive mode. They are also marginally faster than the $ accessor

Issue #424, opened 2020-07-08
SEIR Module Extension -- Fix Errors


smjenness:
@andsv2 : the SEIRModuleExtension.R does not run without error. 

Here is the error I get after running netsim on line 102:

```r
> sim <- netsim(est, param, init, control)

Starting Network Simulation...
 Error in do.call(control[["prevalence.FUN"]], list(dat, at = 1)) : 
  'what' must be a function or character string 
```

It appears that prevalence.net is not getting set as the default function to prevalence.FUN. Why not?

smjenness:
The `SIRwithVitalDynamics.R` file also does not run without error. It is missing an `r.num` initial condition.

smjenness:
Also in the worked examples in the Rmd and accompanying R script, the SI model with no vital dynamics has the `resimulate.network` control setting set to TRUE (why would network resimulation be needed?). And the SIR model *with* vital dynamics has the `resimulate.network` setting set to FALSE (whereas here it would be needed!). Could you explain the thinking here, and correct as necessary.

smjenness:
Similarly, see the first extension example in the JSS2 script (the Age-Dependent Mortality Extension model). Why do you specify the prevalence and network resimulation functions here?

```r
control <- control.net(type = NULL, nsims = 5, nsteps = 500,
                       departures.FUN = dfunc, arrivals.FUN = afunc,
                       prevalence.FUN = prevalence.net, infection.FUN = infection.net,
                       resim_nets.FUN = resim_nets, aging.FUN = aging)
```

smjenness:
Also, with that first extension example, the EpiModel 1.x iteration in the JSS1 script used the `depend = TRUE` control setting to make sure that the network was resimulated. You do not have that specified correctly now with `resimulate.network` because the default value is FALSE.

smjenness:
back to the default network resim and prevalence modules, here's the relevant text in the migration doc:

> Standard modules (in red) are typically not meant to be modified by the user in most applied extension cases. Those modules of class flexible (in blue) are often modified based on novel research cases and are typically related to disease modeling. For built-in models, the protected modules have hard-coded default functions in `control.net`, while the flexible modules have no hard-coded default functions because the appropriate function is selected based on the input parameters from `param.net` and initial conditions in `init.net`.

[...]

PR #425, opened 2020-07-08
Update to EpiModel2 Examples

andsv2:
Fixes for various issues raised in #424.


inst/EpiModel2/SEIRModuleExtension.R
```diff
@@ -96,7 +96,7 @@ param <- param.net(inf.prob = 0.5, act.rate = 2, ei.rate = 0.01, ir.rate = 0.005
 init <- init.net(i.num = 10)
 ## If any user defined functions, type must equal NULL
 control <- control.net(type = NULL, nsteps = 100, nsims = 1, infection.FUN = infect2,
-                       progress.FUN = progress)
+                       progress.FUN = progress, prevalence.FUN = prevalence.net)
```
smjenness, line 103: But prevalence is a module that is in the special class such that it does not need to be specified explicitly, right? Check what is in your migration doc. Shouldn't it be handled like the network resimulation module?

smjenness: @andsv2 This prevalence fx still explicitly defined here. Why?

PR #432, opened 2020-07-23
Major Reworking of Base Colors

smjenness:
* Removed `transco` function (duplication of built-in `adjustcolors` function)
* Remove use of `brewer.pal` for built-in stochastic (ICM, network) models, replaced with base R numerical colors
* Remove duplicate colors with different line types for two-group models --> now different colors with same line type


smjenness:
What do you mean by "style the legend based on color"?

andsv2:
> What do you mean by "style the legend based on color"?

@smjenness Each plotted epitracker would have it's name in the legend the same color as the quant and mean color.

Issue #431, opened 2020-07-23
print.netsim should print model functions used in simulation


smjenness:
This is the built-in two-group model in the `netsim` function documentation:

```r
nw <- network_initialize(n = 100)
nw <- set_vertex_attribute(nw, "group", rep(1:2, each = 50))
formation <- ~edges
target.stats <- 50
coef.diss <- dissolution_coefs(dissolution = ~offset(edges), duration = 20)
est1 <- netest(nw, formation, target.stats, coef.diss, verbose = FALSE)

# Epidemic model
param <- param.net(inf.prob = 0.3, inf.prob.g2 = 0.15)
init <- init.net(i.num = 10, i.num.g2 = 10)
control <- control.net(type = "SI", nsteps = 100, nsims = 5, verbose.int = 0)
mod1 <- netsim(est1, param, init, control)
```

Printing out `mod1` shows the following:
```r
> mod1
EpiModel Simulation
=======================
Model class: netsim

Simulation Summary
-----------------------
Model type: SI
No. simulations: 5
No. time steps: 100
No. NW groups: 2

Model Parameters
-----------------------
inf.prob = 0.3
inf.prob.g2 = 0.15
act.rate = 1

Model Functions
-----------------------
arrivals.FUN = arrivals.net 
departures.FUN = departures.net 
infection.FUN = infection.net 
prevalence.FUN = prevalence.net 

Model Output
-----------------------
Variables: s.num i.num num s.num.g2 i.num.g2 num.g2 
si.flow si.flow.g2
Networks: sim1 ... sim5
Transmissions: sim1 ... sim5
```

Under Model Functions, it is listing modules that are not used in this model (arrivals, departures) and the function names are not correct (these should be the 2g versions). Can this be fixed relatively easily?

PR #433, opened 2020-07-23
Update to print.netsim.

andsv2:
Output from `print.netsim` has been updated to be more flexible and account for base and user extended modules. Previous versions of pulling function names was largely hard coded; this has been completely removed. Now, `f.args` and `f.names` are pulled during `control.net` and updated during `crosscheck.net`. If any functions are missing (NA in `f.args`) these are dropped, maintaining those functions present in control inputs. Note: any user defined function names are parsed as `function` off of `function.FUN`.

Addresses #431 .


smjenness:
@andsv2 : great work, thanks!

PR #446, opened 2020-08-10
Trycatchmodules

AdrienLeGuillou:
Simple wrapping of the module calls that print the name of the module before the warning / error. It will remove the complexity of trying to figure out in which module an error happens.


smjenness:
@AdrienLeGuillou This is great idea. Can you include a MRE in a comment below demonstrating this functionality. Also, please run a basic performance test to ensure this does not add to the runtime.

AdrienLeGuillou:
I switched from `tryCatch` to `withCallingHandlers`, the latter being able to handle warnings without interrupting the function. 

Also, `withCallingHandlers` is now called only once as the full loop is happening inside of it. I ran `microbenchmarks` on my `EpiModelHIV` San Francisco modules, the variance in the module run time was so high compared to the cost of `withCallingHandlers` that the mean and median time were similar and even often lower in the `withCallingHandlers` case.

Example using the test-net-long.R

```r
nw <- network_initialize(n = 100)
est <- netest(nw, formation = ~edges, target.stats = 25,
coef.diss = dissolution_coefs(~offset(edges), 10, 0),
verbose = FALSE)

## "SI, 1M, CL: 1 sim"
param <- param.net(inf.prob = 0.5)
init <- init.net(i.num = 1)
control <- control.net(type = "SI", nsims = 1, nsteps = 25, verbose = FALSE)
x <- netsim(est, param, init, control)

# Addition of a module generating errors and warning
error_module <- function(dat, at) {
  if (at > 19) {
    warning("this is a warning")
  }

  if (at > 20) {
    stop("This is an error")
  }

  dat
}

control <- control.net(type = NULL, nsims = 1, nsteps = 25, verbose = FALSE,
                       error_module.FUN = error_module)
x <- netsim(est, param, init, control)
```

output:

```
A WARNING occured in module 'error_module.FUN': 
	simpleWarning in (function (dat, at) : this is a warning

A WARNING occured in module 'error_module.FUN': 
	simpleWarning in (function (dat, at) : this is a warning

An ERROR occured in module 'error_module.FUN': 
	Error in (function (dat, at) : This is an error

Error in (function (dat, at)  : This is an error
In addition: Warning messages:
1: In (function (dat, at)  : this is a warning
2: In (function (dat, at)  : this is a warning

```

You can suppress these verbose output using:

```
x <- suppressWarnings(suppressMessages(netsim(est, param, init, control)))
```

output

```
Error in (function (dat, at)  : This is an error
```

NOTE: with `withCallingHandlers` keep the original `warning`, `error`, and `message`. That why we see a duplication. I think it does not matter and is better to keep this way as the alternative would be not to manage the warnings and error with `tryCatch`.

AdrienLeGuillou:
@smjenness, I included the modification we discussed. The code is ready for your review

PR #472, opened 2020-11-03
Built in modules need to be flexible when using them in API extension

### LINKED ISSUE #445: departures.net cannot handle type = NULL for custom modules
I am using some custom modules and some base modules. Given that now `type = NULL` must be specified when calling non-base models I've got the following error:

```R
Error in if (type == "SIR") { : argument is of length zero  
```
Which comes from the `EpiModel::departures.net` function. This occurs because if `type` is NULL, `NULL == 'SIR'` evaluates to logical(0), which cannot be handle. I think you need to call `isnull()` and then handle non standard models appropriately.

My current `control.net` call is as follows:
```R
  control = EpiModel::control.net(
    type = NULL,
    nsims = nsims, 
    nsteps = nsteps,
    # delete.nodes = T,  this does not work for now
    ncores = cores,
    initialize.FUN = custom.initialize.net, # this bit is just so that I can extract time
    exposure.FUN = exposure,
    infect.FUN = infect,
    epi.by = "age",
    quarantine.FUN = quarantining,
    hospitalize.FUN = RequireHospitalization,
    recover.FUN = recover,
    fatality.FUN = fatality,
    departures.FUN = EpiModel::departures.net,
    prevalence.FUN = custom.get_prev.net,
    skip_check = FALSE,
    resimulate.network = T,
    module.order = c('exposure.FUN', 'infect.FUN',
                     'quarantine.FUN', 'hospitalize.FUN', 'recover.FUN',
                     'fatality.FUN', 'departures.FUN', 'prevalence.FUN')
  )
```

All the functions in module.order have `(dat, at)` as params. I'll fork the repo and submit a PR which could satisfy you. Which things need to be handle if `type = NULL` that would be different from `type %in% c('SIR', 'SI')`?

smjenness:
Closes #445 

Work in Progress...


AdrienLeGuillou:
I fixed the simpler issue in #571 to start closer from the current master and opened a broader discussion on #572

smjenness:
Closing this PR without merging based on ALG comment above.

PR #484, opened 2020-12-03
Rng params

AdrienLeGuillou:
Implementation of random parameters mechanism for `param.net` and `netsim
`


smjenness (changes requested):
Longer comment below

smjenness:
Nevermind, there was a typo in the example. Things are working as expected.

smjenness (approved):

Issue #485, opened 2020-12-07
Add a UID attribute to network nodes


AdrienLeGuillou:
related to these two issues: #478 #482 

the goal would be to have a Unique Identifier (uid) attribute for each node in the network. 

it can be stored either as an `integer` (up to 4 billions unique values, negatives and positive) or `double` (way more values but larger memory requirements)

it would probably be beneficial to create a function for adding new nodes that ensures every attributes required by core EpiModel are set (`active` and `uid`).  This function would in turn be available for modules creators to ensure the compatibility with core EpiModel machinery.

smjenness:
> it would probably be beneficial to create a function for adding new nodes that ensures every attributes required by core EpiModel are set (active and uid). This function would in turn be available for modules creators to ensure the compatibility with core EpiModel machinery.

Related to this, it would be good also to standardize the writing and appending of all the core attributes needed for open population models: http://statnet.org/tut/EpiModel2.html#extending-arrivals-departures

smjenness:
Closed with #488

Issue #482, opened 2020-11-24
Transmission Matrix Not Converting to Phylo Object


thednainus:
Hello, 
I'm trying to write some code for HIV dynamics and I tried to save the transmat. However, when I try to convert the transmat to a phylogenetic tree using as.phylo.transmat, I seem to be going to a forever loop. 
I tried the example for covid here (http://statnet.org/nme/d5-s4-COVIDDemog.html) just to make sure it was not anything that I did not code correctly. However, when I get the transmat, I also end up into a loop that never ends for some values of v_sub (in the function as.phylo.transmat). Do you know what might be the cause?

Thanks,
Fabricia.

smjenness:
Also tagging @skyebend here, as he was the one to originally develop this conversion code, several years ago. Skye if you have any quick impressions, they would be appreciated.

thednainus:
Thanks for transferring this to EpiModel; I thought I posted that to the right R package (not sleeping much as my toddler is keeping me up at night)... I am new to this type of network modelling. For me to try to find a solution, I need to understand why a loop is being generated... Is this because of a cyclic graph?

thednainus:
I've been trying to fix this bug. It might be something else that is causing it and _maybe_ not something wrong with the as.phylo.transmat function .

I did a few tests, and I realized that sometimes a same susceptible gets infected more than once in different time points. If I understood the code correctly that is not possible, right? Because we are looking for pairs of _infected_ and _not infected_ for a transmission to happen.

An example of a transmission matrix in which show a same susceptible getting infected more than once in different time points. Do you think that this could be the cause? If we don't expect this to happen, where exactly should I be looking to fix it?
I did check whether 1 was being assigned to newly infected individuals (and that seems to be the case, or maybe I am missing something).

```
     **at    sus     inf     transProb    actRate    finalProb**
107     575    652       0.1              3              0.271
97      575     913       0.1             3               0.271 
```

thednainus:
I think I found the bug. Unfortunately, I just don't think I will have the time to update EpiModel and do a pull request.
I think the actual problem is in the function `[discord_edgelist](https://github.com/statnet/EpiModel/blob/dc3640500dc0b6a1fb9f7dc8a68d73490abcfe23/R/net.mod.infection.R#L303)`.
What is happening is that everything related to susceptible and infected should be using index as in for example:

[...]

Issue #368, opened 2020-04-09
the `dat` master list object for network simulation should contain the `at` element


AdrienLeGuillou:
In network simulation, we pass around `dat` and `at` to each module. I think it would be useful to have `at` be a part of the `dat` object.

This would push the `dat` object to be closer to a state of "sufficient and necessary" for network simulation. That would simplify the restarting process for example.

It would allow getters and setters (for `epi` in particular) to be aware of the step the simulation is in at any point and permits functions like `append_epi(dat, tracker, value)` to add Epi Outcome values at the right spot.

The changes required in code are minimal:
- set `dat$at <- 1` in the initialize module
- `dat$at <- dat$at + 1` at the end of the "TIME LOOP" in the `netsim_loop` function

what do you think?

smjenness:
Interesting idea. I think I like it. But let's put it on the back burner for now.

PR #488, opened 2020-12-15
Implement General Framework for Unique IDs in EpiModel

AdrienLeGuillou:
WIP

This PR will add the "uid" attribute to the core EpiModel network functionalities. It intend to fix issues #485 and #482 

ToDo

- [X] implementation
    - [X] UID attributes
    - [X] use of UID in `set_transmat`  
- [x] unit tests
- [X] documentation


R/net.accessor.R
```diff
@@ -18,14 +17,22 @@
 #' @return a vector or a list of vector for `get_` functions. And the Master
 #'         list object for `set_` and `add_` functions
 #'
+#' @section Core Attribute:
+#' The `append_core_attr` function initialize the attributes necessary for
```
smjenness, line 21: The documentation is not built with markdown currently (I was getting an error when I tried to convert, I'll deal with a conversation later), so please change `x` to \code{x}

AdrienLeGuillou: Oh ok, I updated that

AdrienLeGuillou:
Ready for review: 

Summary of the changes

1. one new function available to module creators: `append_core_attr()`. This function initiate the attributes of new nodes required by the rest of the EpiModel machinery (active and uid). 

2. modification of the `append_attr()` function. It can now create non existent attributes.  This makes the initial creation of nodes and the addition of new nodes in open population models more consistents. Also, it now uses the `get_attr` and `set_attr` functions internally and not specific code.

3. modification of the `set_transmat` function. It now uses UIDs instead of indexes to reference nodes

4. various corrections of bugs / typos

smjenness:
@AdrienLeGuillou : this is very exciting, thank you! 

Questions before I approve/merge:
1. did you test this against the user's example in #482? They were running the NME COVID model (linked in the issue) and then outputting the transmat to a phylo, and it was breaking there. 
2. I think it would be nice to see an easy demonstration of how this works with an update to this example: https://github.com/EpiModel/EpiModel-Gallery/tree/master/2018-08-SIwithVitalDynamics. Ultimately, we will need to go through all of the gallery and tutorial examples to use this new functionality, but that's a good starting point. 
3. We'll want to use this for EpiModelHIV-p/CombPrevNet, to replace the hacky version of UID there. can you just quickly scan through the code as you are doing your validation there to see if there will be any potential problem with that replacement; and then once this is merged I'll ask you to implement that replacement.

AdrienLeGuillou:
1. I have, I used these two files: 

- http://statnet.org/nme/d5-s4-COVIDDemog.R where I used this branch of EpiModel and added at the end

``` r
tm <- get_transmat(sim)
phylotm <- as.phylo.transmat(tm)

plot(phylotm) # will plot 38 plots
```

- http://statnet.org/nme/d5-s4-COVIDDemog-fx.R where I changed 
`dat <- append_attr(dat, item = "active", value = 1, n.new = nArrivals)` into `dat <- append_core_attr(dat, nArrivals)`

once this is approved, I could detail the necessary changes on the issue #482 

2. I made this PR implementing it: https://github.com/EpiModel/EpiModel-Gallery/pull/30 - I can make the rest of the updates pretty quickly if you want (30 min to an hour including testing) or I can wait

3. I did the update in this commit https://github.com/EpiModel/EpiModelHIV-p/commit/c69b01acacfe9bd6f81df4d7e13ecfa2f7377cf3 (the next one is just a formatting mistake) - I kept the old code in comment as I have not tested the code yet

smjenness:
Thanks @AdrienLeGuillou 

1. I think `append_core_attr` should also be auto-writing the other two required attributes: entrTime and exitTime. See http://statnet.org/tut/EpiModel2.html#extending-arrivals-departures. Then in the examples, it is only necessary to write append_core_attr, append_attr for status (since this may vary from model to model, it should not be in append_core_attr), and anything else bespoke. 
2. Could you move this new function out of a separate R file and into the main net accessor R file (I would like all of these utility functions together).

R/net.core.attr.R
```diff
@@ -0,0 +1,61 @@
...
+#' the /code{dat[["_last_uid"]]} variable
+#'
+#' @param dat a Master list object of network models
+#' @param n.new the number of new nodes to give \code{uid} to
+#'
+#' @return the Master list object of network models (\code{dat})
+#'
+#' @keywords internal
+update_uids <- function(dat, n.new) {
+  last_uid <- if (is.null(dat[["_last_uid"]])) 0L else dat[["_last_uid"]]
```
smjenness, line 24: Is this what initializes the `dat[["_last_uid"]]`? If not, what does? If so, shouldn't this be initialized at the current network size (e.g., `sum(dat$attr$active == 1)`) instead of `0L`?

AdrienLeGuillou:
> Thanks @AdrienLeGuillou
> 
>     1. I think `append_core_attr` should also be auto-writing the other two required attributes: entrTime and exitTime. See http://statnet.org/tut/EpiModel2.html#extending-arrivals-departures. Then in the examples, it is only necessary to write append_core_attr, append_attr for status (since this may vary from model to model, it should not be in append_core_attr), and anything else bespoke.
> 
>     2. Could you move this new function out of a separate R file and into the main net accessor R file (I would like all of these utility functions together).

I agree for both, I will do the modifications. Should there also be a similar function to handle these attributes for departures?

smjenness:
> I agree for both, I will do the modifications. Should there also be a similar function to handle these attributes for departures?

Do you mean an auto-update accessor function specifically for the active and exitTime attributes?

R/net.core.attr.R
AdrienLeGuillou: It is, if `dat[["_last_uid"]]` does not exist it sets it to 0L (enforcing integer to reduce the memoy footprint)

On the first call of `append_core_attr` in the initialization module, there are no nodes yet. `n.new` are going to be created. So when `update_uids` is called for the first time, `dat[["_last_uid"]]` must be 0L as no UIDs have been used yet.

The next line creates the first `n.new` UIDs with `seq_len(n.new) + last_uid` (1:n.new in this case as `last_uid == 0L`.

`dat[["_last_uid"]]` is then updated to reflect the fact that UIDs from 1 to n.new are used

AdrienLeGuillou:
> > I agree for both, I will do the modifications. Should there also be a similar function to handle these attributes for departures?
> 
> Do you mean an auto-update accessor function specifically for the active and exitTime attributes?

yes, mainly to be consistent in the handling of core attributes. This would also make it easier if at some point other attributes become necessary.

This way we would have `append_core_attr()` en initialization and arrival and the new function (can't think of a good name right now) in departure.

smjenness:
I think it's ok to leave as-is for now -- the specific calculation of exitTime is wrapped up in the broader departure function, which makes generalization tricky, and the burden is relatively low here (calling set_attr two times).

AdrienLeGuillou:
Ok, I just wanted to be sure

R/net.core.attr.R
smjenness: makes sense now, thanks

AdrienLeGuillou:
Changes made and checked with #482 

NB: on CombPrevNet entrTime is called arrival.time. So with the update there are both entrTime and arrival.time

I had to add an `at` argument to `append_core_attr`. At some point (not now) think I should revisit #368 as it would simplify a lot of the code and the creation of modules

smjenness:
> NB: on CombPrevNet entrTime is called arrival.time. So with the update there are both entrTime and arrival.time

For EpiModelHIV, please replace arrival.time with entrTime, to avoid this duplication

AdrienLeGuillou:
> > NB: on CombPrevNet entrTime is called arrival.time. So with the update there are both entrTime and arrival.time
> 
> For EpiModelHIV, please replace arrival.time with entrTime, to avoid this duplication

I will when I start running the combprevnet code. I don't want to break anything by changing it without running it

smjenness (approved):

PR #493, opened 2021-01-04
Implement new approach to calculate relational age in netdx

sgoodreau:


sgoodreau:
In NMG yesterday, we discussed how to make this work I need to change a detail in the calculation of mean age of extant ties from an edgelist.  That is, right now the calculation occurs for every time in the range ``min(terminus):max(terminus)``, which was fine when it was just printing, but creates problems when I need to calculate a global mean.  We discussed switching to `1:max(terminus)`` which folks were fine with, because ``netdx`` objects always start at time 1 nd don't use any kind of resimulation functionality.

Turns out that when I said I only need to change ``net.dx``, I was wrong -- this calculation occurs in the utility function ``edgelist_meanage``.  If anyone else calls that function, then a problem could ensue.  I have searched through the code base on ``EpiModel`` and ``EpiModelHIV`` and I do not see it used anywhere else.  It is not a function people are likely to use uch on their own, but it is exported and documented (albeit not clearly).

Still OK for me to edit this function?  @smjenness, @martinamorris, @dth2 ?

smjenness:
Yes, please edit @sgoodreau

dth2:
I am ok with the change

martinamorris:
@adam-s-elder do you use this in any of your calculations?

adam-s-elder:
> @adam-s-elder do you use this in any of your calculations?

I don't think so.

smjenness:
@sgoodreau is this ready to go once you fix the `dgeom` import?

sgoodreau:
alas, no.  the print.netdx bit is done, but i want to now align the plot.netdx to reflect the new ideas, and then edit the documentation for both.  what's your current timeline for releasing a new version to CRAN?

smjenness:
@sgoodreau : no rush if it's not done and you need some time. Let's set a target date of February 18 (2 weeks from today). I will update the v2.0.4 accordingly.

sgoodreau:
@smjenness  I'm curious about the following code in ``plot.netdx`` that appears around line 1626:

```
if (dynamic == TRUE) {
      if (is.numeric(qnts) & nsims > 1) {
        if (qnts < 0 | qnts > 1) {
          stop("qnts must be between 0 and 1", call. = FALSE)
        }
        if (missing(qnts.col)) {
          qnts.col <- sim.col
        }
        if (missing(qnts.alpha)) {
          qnts.alpha <- 0.5
        }
        qnts.col <- adjustcolor(qnts.col, qnts.alpha)
        quants <- c((1 - qnts) / 2, 1 - ((1 - qnts) / 2))
        dataj <- as.data.frame(pages)
        dataj <- dataj[complete.cases(dataj), , drop = FALSE]
        qnt.prev <- apply(dataj, 1, function(x)
                          quantile(x, c(quants[1], quants[2]),
                                   na.rm = TRUE))
        xx <- c(1:(ncol(qnt.prev)), (ncol(qnt.prev)):1)
        if (qnts.smooth == FALSE) {
          yy <- c(qnt.prev[1, ], rev(qnt.prev[2, ]))
        } else {
          yy <- c(suppressWarnings(supsmu(x = 1:(ncol(qnt.prev)),
                                          y = qnt.prev[1, ]))$y,
                  rev(suppressWarnings(supsmu(x = 1:(ncol(qnt.prev)),
                                              y = qnt.prev[2, ]))$y))
        }
```

This is called under the condition ``type == "duration" & method == "l" & dynamic == TRUE & is.numeric(qnts) & nsims > 1``.  But the very similar code (minus a few lines) is called earlier in the function under a less restrictive condition (``type == "duration" & dynamic == TRUE & is.numeric(qnts) & nsims > 1``).  So if it is to be run the second time, then it's always already been run the first time.  I think what happened is that the second time was written first (for the purposes of plotting the polygon) and then you realized that you needed it earlier and more broadly for the purposes of setting the ylim, and added it there.

In short, the code works as is, but I think can be simplified by removing most of the second copy of this specific code.  I don't know your logic nearly as well as you do though, so just want to comment on it and leave it for you to decide whether to add an issue.  If so, I'm happy to pull it out and test it.

sgoodreau:
OK the current version has both the printing and plotting updated. You can try it out with the code below.  At this point, 
I'd like input from NMG (@martinamorris, @dth2, @emilypo, @adam-s-elder, and especially @smjenness) about new plotting arguments.  Right now, it will plot both the observed and the imputed versions of whichever elements (sim lines, lean lines, quantiles) you ask it to plot.  And do them in the same color as each other.  It seems to me it would be good to have new arguments asking whether you want the observed values (default==TRUE) separately from the imputed values (default also ==TRUE).  No need to allow for every combination for each of the three elements, just across the board. How to handle colors is more complicated, as the arguments quickly get numerous. Similar for smoothing.  Something parsimonious but with the most important components of flexibility would be good.  Preferences?

```
#library(devtools)
#library(testthat)
#library(EpiModel)

n <- 200
my.net <- network.initialize(n, FALSE)
handedness <- rep("R", 200)
handedness[sample(1:n, round(n*0.1), replace = FALSE)] <- "L" 
my.net <- set.vertex.attribute(my.net, "handedness", handedness)

my.nsteps <- 2e2
my.num.sims <- 5
  
my.formation.1 <- ~edges
my.target.stats.1 <- c(150)

my.formation.2 <- ~edges+nodematch("handedness")+degree(0)
my.target.stats.2 <- c(150, 125, 20)

my.duration.1 <- 25
my.coeff.diss.1 <- dissolution_coefs(~offset(edges), my.duration.1)

my.duration.2 <- c(25, 10)
my.coeff.diss.2 <- dissolution_coefs(~offset(edges)+offset(nodematch("handedness")), my.duration.2)

my.fit.11 <- netest(my.net, formation = my.formation.1, target.stats = my.target.stats.1,
                   coef.diss = my.coeff.diss.1, edapprox = TRUE)
my.fit.21 <- netest(my.net, formation = my.formation.2, target.stats = my.target.stats.2,
                    coef.diss = my.coeff.diss.1, edapprox = TRUE)
my.fit.22 <- netest(my.net, formation = my.formation.2, target.stats = my.target.stats.2,
                    coef.diss = my.coeff.diss.2, edapprox = TRUE)

my.netdx.11 <- netdx(my.fit.11, nsteps = my.nsteps, nsims = my.num.sims)
my.netdx.21 <- netdx(my.fit.21, nsteps = my.nsteps, nsims = my.num.sims)
my.netdx.22 <- netdx(my.fit.22, nsteps = my.nsteps, nsims = my.num.sims)

my.netdx.111 <- netdx(my.fit.11, nsteps = my.nsteps, nsims = 1)
my.netdx.211 <- netdx(my.fit.21, nsteps = my.nsteps, nsims = 1)
my.netdx.221 <- netdx(my.fit.22, nsteps = my.nsteps, nsims = 1)


####### revised print.netdx output 

my.netdx.11
my.netdx.21
my.netdx.22

my.netdx.111
my.netdx.211
my.netdx.221

####### revised plot.netdx output 

# plot(my.netdx.21, type = "duration") Not needed, same as my.netdx.11
# plot(my.netdx.22, type = "duration") Returns an error by design (for now)

# multi-sim case; a couple different versions to make sure all the components work

plot(my.netdx.11, type = "duration")

plot(my.netdx.11, type = "duration", sim.lines = T, qnts = 1, 
     mean.line = T, targ.line = T, mean.smooth = F, qnts.smooth = F)

plot(my.netdx.11, type = "duration", sim.lines = T, qnts = 0.5, 
     mean.line = T, mean.smooth = T, targ.line = T, qnts.smooth = T)

# single-sim case; a couple different versions to make sure all the components work

plot(my.netdx.111, type = "duration")

plot(my.netdx.111, type = "duration", sim.lines = T, qnts = 1, 
     mean.line = T, targ.line = T, mean.smooth = F, qnts.smooth = F)

plot(my.netdx.111, type = "duration", sim.lines = T, qnts = 0.5, 
     mean.line = T, mean.smooth = T, targ.line = T, qnts.smooth = T)
```

sgoodreau:
As per convo in NMG today, we will set it so that ``plot.netdx`` plots *either* the observed or the imputed relational ages, not both, with the imputed as the default.

smjenness:
@sgoodreau : Regarding the parameter name, I think it should be a logical (T/F) called `duration.imputed` with default of TRUE.

R/plot.R
```diff
@@ -1523,8 +1535,17 @@ plot.netdx <- function(x, type = "formation", method = "l", sims, stats,
            currently available", call. = FALSE)
     }
 
+    if (missing(duration.imputed)) {
+      duration.imputed <- TRUE
+    }
```
smjenness, line 1540: If this has a universal default, it can be specified directly in the parameter list above instead of like this.

sgoodreau: Done.

smjenness (approved):

smjenness:
thanks @sgoodreau !

smjenness:
@sgoodreau : I did not notice until now but there is a bug introduced with this on print.netdx. In your example above, this is what the print looks like:

```r
my.netdx.221
EpiModel Network Diagnostics
=======================
Diagnostic Method: Dynamic
Simulations: 1
Time Steps per Sim: 200

Formation Diagnostics
----------------------- 
  Target Sim.Mean Pct.Diff Sim.SD
1    150   142.51    -4.99     NA
2    125   116.27    -6.98     NA
3     20    25.58    27.87     NA

Dissolution Diagnostics
----------------------- 
  Target Sim.Mean Pct.Diff Sim.SD
1  25.00    13.49   -46.04     NA
2   0.04     0.09   120.96     NA
----------------------- 
* Heterogeneous dissolution model results averaged over
```

The row names of both tables were removed, and now only show row numbers. You probably deleted something within netdx that broke this. We should probably have a unit test that checks against the proper output too, if you could add.

PR #505, opened 2021-02-18
Restructure the lines that round the stats.table.formation and statst…

sgoodreau:
…able.dissolution so as not to lose the row names.


smjenness:
Fixing this on new relational_age2 branch

PR #506, opened 2021-02-19
issue 491: set_attr should also have indexes

### LINKED ISSUE #491: set_attr should also have indexes arg
similar to get_attr. 

so we can do this in line:

```
attr <- get_attr(dat, "attr")
toUpdate <- 2:3
set_attr(dat, "attr", value = c(99, 99), indexes = toUpdate)
```

AdrienLeGuillou:
fixes #491 

This pull requests adds an `indexes` arguments to `set_attr`, similar to the one for `get_attr`. 

The default value of this parameter is `NULL` so when `set_attr` is used without this attribute, it behaves exactly as before.

The function itself now call one of two non exported subfunction:

- `set_attr_full` for when `indexes = NULL`
- `set_attr_indexeds` otherwise

this split was made because they have a different set of error checks and the readability was becoming hard in one function.

Tests have been written to check theses new functions


AdrienLeGuillou:
After more checks (and error fixing) I re-merged `set_attr` into a single function. The main reason is to simplify the error messages. With the 2 sub-functions, the error messages are displayed with the name of the sub-function. Which makes the programs a bit harder to debug

smjenness (approved):

smjenness:
Looks great, thanks @AdrienLeGuillou

PR #510, opened 2021-03-02
Update dissolution

chad-klumb:
Currently there are a few issues with `update_dissolution`:

- it was missed when we went from `log(D - 1)` to `log(D)` in the EDA formation correction and thus is out of date with how `netest` behaves
- this should have been caught by the `update_dissolution` test, but it appears that there are misplaced parentheses in that test, causing it not to function as intended
- currently, `update_dissolution` assumes that the old and new dissolution models have the same number of coefficients; in general, this need not be true (e.g. we could go between homogeneous and heterogeneous dissolution)

This PR fixes the above issues and adds a new test for switching between homogeneous and heterogeneous dissolution.  It is assumed (as always, right now) that the relevant dissolution models occur at the beginning of the ergm model.

### DIFF
--- a/R/netest.R
+++ b/R/netest.R
@@ -366,9 +366,12 @@ update_dissolution <- function(old.netest, new.coef.diss) {
 
   out <- old.netest
 
-  l.cd <- length(new.coef.diss$coef.crude)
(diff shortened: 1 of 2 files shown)


smjenness:
Thanks @chad-klumb

PR #512, opened 2021-03-10
non-nested EDA

chad-klumb:
Closes statnet/EpiModel#501

Adds arguments and logic to `netest` and `update_dissolution` to handle non-nested EDA applications.

The default remains the old (nested) form.

Related tests were also added.

A technical point may arise regarding formula environments, if the dissolution formula environment contains required variables that are not also present in the formation formula environment.  In that case, the `from.new` argument of `nonsimp_update.formula` can be passed through `netest` and `update_dissolution` via the `...` to achieve the required copying (see the new tests for some examples).  For the time being I am (intentionally) not calling a lot of attention to this; if it turns out to be a frequent cause of consternation we can bring it more to the forefront.


smjenness:
Thanks @chad-klumb !!!

Issue #386, opened 2020-04-19
Allow for saving full attribute list not truncated with deaths/departures


smjenness:
Making this consistent with EvoNet. Could probably be a post-2.0 addition.

AdrienLeGuillou:
What exactly do we want to save?

If it's all the attributes at the time of exit (or the last step of simulation),
the implementation is pretty easy, we simply need to store the `attr` for the
"dead" nodes in the departure module, and append the `attr` list at the end of
the simulation.

From what I understand, evoNet is using this to recover some value like "first
time on treatment" and not really the state of the moving variables (like age or
CD4) at the last time step.

If I understand correctly, we might want to think of a way to store some
specific elements in a format that allows more flexibility and uses less space.
For example a JSON like structure:

``` r
list(
  node1 = list(),
  node2 = list(
    tx.starts = c(20, 700),
    tx.stops = c(70, 1000)
  ),
  node3 = list(
    tx.starts = c(10, 100),
    tx.stops = c(70),
    prep.start = c(1),
    prep.stop = c(9)
  ),
  node4 = list()
)
```

a lot of different options are possible and I think we should discuss mainly the
use cases to find the best approach

PR #515, opened 2021-03-31
add get_uids and get_indexes functions

AdrienLeGuillou:
This PR would add two functions to EpiModel

`get_uids` : that get a vector of indexes and return the corresponding node UIDs
`get_indexes` : that get a vector of nodes UIDs and return the corresponding indexes

this back and forth between the two can become quite common, especially with time varying attributes #386. A predefined implementation will allow better checks and a stable API should the implementation change.


smjenness (changes requested):
Just the one comment below about posit_ids for epi. Otherwise, this looks good!

R/net.accessor.R
```diff
@@ -260,7 +259,7 @@ get_epi_list <- function(dat, item = NULL) {
 
 #' @rdname net-accessor
 #' @export
-get_epi <- function(dat, item, indexes = NULL, override.null.error = FALSE) {
+get_epi <- function(dat, item, posit_ids = NULL, override.null.error = FALSE) {
```
smjenness, line 262: posit_ids does not have the same meaning for epi as it does for attr, because the time step in epi is not really an ID conceptually but just a marker for the time step, at. Can we just use `at` for the argument name?

smjenness (approved):

PR #518, opened 2021-04-09
remove foreach / dopar (fixes issue 517)

### LINKED ISSUE #517: Change built-in parallelization functions from doParallel/foreach to parallel/parLapply
Enforce a default "PSOCK" type of parallel::makeCluster to make built-in parallel set up across all systems

AdrienLeGuillou:
fixes #517


AdrienLeGuillou:
@smjenness 
 I looked at the issue a bit more and I think the best option for now is to keep `foreach`.

Switching to native "PSOCK" clusters replicate the windows specific issues on all platform and add a few more. Whith `foreach` and `dopar`, these packages will try to infer from the `%dopar%` code what is required by the parallel processes. It works in most cases but sometimes the objects are hidden inside others and that cause the errors. With `parallel`, nothing is infered, the package assumes that the best practice of parallel computing are followed, namely: everything that is required is passed as argument to the parallel function. 

Because EpiModel was built around `%dopar%`, getting rid of it would mean rethinking all the architecture of the package. Same thing for fixing the errors on windows. 

I believe that it would not be worth it, especially because most of these errors are with very complex models that would run on linux clusters where the issue is not present.

martinamorris:
I haven't followed all of the background here, but since we have gotten questions from users on parallel use cases, for Windows in particular, I think it would be helpful to have a wiki on using parallel for EpiModel in different contexts -- what works, and what doesn't.

smjenness:
We are going to close this without merging per discussion, because it would require a major overhaul of parallel architecture...

PR #526, opened 2021-05-13
get all unique / posit IDs

AdrienLeGuillou:
Small complement to the `get_unique_ids` /  `get_posit_ids` functions.

now if the argument `unique_ids` or `posit_ids` is NULL (default value) the functions will return the full vector of positional IDs or unique IDs respectively


smjenness (approved):

smjenness:
@AdrienLeGuillou makes sense, looks good, thanks!

Issue #444, opened 2020-08-07
Create a flexible "prevalence module" for netsim


AdrienLeGuillou:
The idea is to allow the user to add "epi trackers" at runtime, without the need to modify the modules themselves.

This idea is also implemented [in this branch of EpiModelHIV-p](https://github.com/EpiModel/EpiModelHIV-p/blob/SFO_PrEP/R/mod.prevalence.R).

The user has to provide as list of function through the `param$epi_funs` parameters. One such function can look like this:

``` r
epi_linked_6month <- function(dat, at) {
  with(dat$attr, {
    cond <-  tx.init.time - diag.time <= 26
    pop <- diag.status == 1 

    sum(cond & pop, na.rm = TRUE) / sum(pop, na.rm = TRUE)
  })
}
```

The module can be simplify compared to EpiModelHIV, by using the new `set_epi` function like so:

``` r
prevalence.net <- function(dat, at) {
  epi_funs <- get_params(dat, epi_funs)

  for (fn in names(epi_funs)) {
    dat <- set_epi(dat, fn, at, epi_funs[[fn]](dat, at) )
  }
}
```

This method allows the use of functional programming magic to create function factories that simplify the creation of a lot of trackers. For instance, if I want the proportion of black people that had a viral load below 200 at any point in the past year I would add this tracker:

``` r
param$epi_funs <- c(
  param$epi_funs, 
  "sup_dx_this_year.B" = epi_sup_dx_this_year(1)
)
```

that use the following function:

``` r
epi_sup_dx_this_year <- function(r_ind) { # r_ind is the race(s) of interest
  # the output of the function is the actual tracker. (Hence the term "function factory") 
  function(dat, at) {
    with(dat$attr, {

      cond <- at - vl.last.supp <= 52
      pop <- (at - diag.time <= 52 + 52 & at - diag.time >= 52 &
                race %in% r_ind)

      sum(cond & pop, na.rm = TRUE) / sum(pop, na.rm = TRUE)
    })
  }
}
```

for the same tracker but without race distinction, I could then use

``` r
param$epi_funs <- c(
  param$epi_funs, 
 "sup_dx_this_year" = epi_sup_dx_this_year(1:3)
)
```

More example of these functions [can be looked at here](https://github.com/EpiModel/EpiModel_SFO/blob/reload/R/utils-prev_funs.R)

AdrienLeGuillou:
[...]

PR #530, opened 2021-05-24
Custom trackers

AdrienLeGuillou:
This will import the custom tracker module made for EpiModelHIV into core EpiModel as an optional module.

This module uses a list of functions stored into `dat$param$epi_trackers` to populate the `epi` list in additions to the trackers built in the `prevalence` modules. This provides a convenient way to add new trackers without modifying the model code.

related to: #444 

To Do:

- [x] Copy the code from EpiModelHIV and document it
- [x] make the names consistent with the core EpiModel style
- [x] make a small vignette describing the use cases


R/net.mod.trackers.R
```diff
@@ -0,0 +1,119 @@
...
+#' tracker.list <- list(
+#'   prop_infected = epi_prop_infected,
+#'   s_num         = epi_s_num
+#' )
+#'
+#'  # Do not forget to add it to `param`
+#'  param <- param.net(
+#'    inf.prob = 0.3,
+#'    act.rate = 0.5,
+#'    tracker.list = tracker.list
```
smjenness, line 63: I feel like this custom tracker is more of a "control" rather than a "parameter", since in effect it is modifying the prevalence module defined in the controls. Feel free to convince me otherwise. :)

smjenness: i.e., it should go under `control.net` rather than `param.net`

AdrienLeGuillou: That was also my initial idea but I was unsure of exactly what defines a "parameter" and a "control". I don't even remember what was my reasoning for choosing `param.net` and not `control.net`. I'll switch

smjenness (approved):

PR #531, opened 2021-05-24
Update the model parameters during runtime

AdrienLeGuillou:
This will import the EpiModel HIV module allowing the model to change parameters  during runtime into core EpiModel as an optional module.

To Do:

- [x] Copy the code from EpiModelHIV and document it
- [x] make the names consistent with the core EpiModel style
- [x] make a small vignette describing the use cases


R/net.mod.updater.R
```diff
@@ -0,0 +1,146 @@
...
+#'    inf.prob = 0.3,
+#'    act.rate = 0.5,
+#'    hiv.test.rate = rep(0.256, 3),
+#'    trans.scale = c(1, 2, 3),
+#'    param.updater.list = param.updater.list
+#'  )
+#'
+#' # Enable the module in `control`
+#'  control <- control.net(
+#'    type = NULL, # must be NULL as we use a custom module
```
smjenness, line 83: because of this, the updater module will be the only one that runs. can you use a simple closed-population SI model to demonstrate instead? For example, you need to define an infection module.

vignettes/Updating-the-model-parameters-during-runtime.Rmd
```diff
@@ -0,0 +1,164 @@
...
+```{r updater-module, echo = FALSE}
+ control <- control.net(
+   type = NULL, # must be NULL as we use a custom module
+   nsims = 1,
+   nsteps = 20,
+   verbose = FALSE,
+   updater.FUN = updater.net
+)
+```
+
```
smjenness, line 108: this should demonstrate a full epidemic model (small population, single sim, limited time steps); perhaps the parameters are different enough at different time points that it is clear to see the impact of the changes on the prevalence or incidence curves

R/net.mod.updater.R
AdrienLeGuillou: done

smjenness (approved):

PR #544, opened 2021-06-07
Merge of Statnet4 dev functionality

smjenness:


R/net.mod.simnet.R
```diff
@@ -110,19 +110,39 @@ resim_nets <- function(dat, at) {
...
     }
+    
+    if (dat$control$save.nwstats == TRUE) {
+      nwL <- networkLite(dat$el[[1]], dat$attr)
+      if (dat$control$tergmLite.track.duration) {
+        nwL %n% "time" <- dat$p[[1]]$state$nw0 %n% "time"
+        nwL %n% "lasttoggle" <- dat$p[[1]]$state$nw0 %n% "lasttoggle"
+      }
+      dat$stats$nwstats[[1]] <- rbind(dat$stats$nwstats[[1]],
+                                        summary(dat$control$nwstats.formulas[[1]],
```
smjenness, line 142: just a note that it is `nwstats.formula` (singular), and not an embedded list (so control$nwstats.formula would be of class formula not list). I think we had decided yesterday to keep it like that for now.

smjenness: see here: https://github.com/statnet/EpiModel/issues/541#issuecomment-856119470

chad-klumb: init_tergmLite organizes some things into lists in anticipation of the general multiple-network case

the user of core EpiModel would specify `nwstats.formula = whatever`, as they do for EpiModel tergm simulation, and not need to worry about these internal details

PR #558, opened 2021-06-30
eliminate low-level reference in tergmLite

chad-klumb:
references statnet/tergmLite#51, statnet/tergmLite#54

I am seeing < 5% mean runtime differences for the netsim call in the EpiModelHIV.R example in tergmLite on both unix and windows (with length of the simulation increased to 50 years)

requires minor updates to EpiModelHIV@CombPrev-dev which I will push after this and statnet/tergmLite#54 are merged

temporarily induces dependence on github version of tergmLite, but we can merge the rest of tergmLite into EpiModel shortly (and move networkLites into a lower package at some point later, if we want to), so that dependence will be transient

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -60,4 +60,4 @@ VignetteBuilder: knitr
 LinkingTo: ergm
 RoxygenNote: 7.1.1
 Encoding: UTF-8
-    
+Remotes: github::statnet/tergmLite@top_level_simulate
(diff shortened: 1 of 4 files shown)


chad-klumb:
Also, I would note that I removed the `simulate` wrappers from `tergmLite` in favor of direct `simulate` calls in `EpiModel`, which don't really complicate things at all and eliminate an intermediate layer of complexity.

smjenness:
Thanks @chad-klumb. I will probably let this work hang out in this branch until after NME (Aug 2 to 6). It's unlikely I will have time to release a new version of these packages on CRAN before then.

chad-klumb:
Okay.  I may stack changes on top of these, in other branches, to address e.g. statnet/tergmLite#52 and statnet/EpiModel#543, but there's no reason any of that needs to be merged into `master` before NME.

PR #559, opened 2021-07-04
remove tergmLite dependency

chad-klumb:
stacked on #558 

references statnet/tergmLite#52

merges what's left of tergmLite into EpiModel, eliminating the need for a separate tergmLite package

mostly tests and documentation with relatively little actual code

we can move networkLites elsewhere at a later date if we chose to


smjenness:
Thanks @chad-klumb. Can you provide a timing benchmark here of the current (CRAN) versus this method. Maybe an EpiModelHIV (10k node, 3 layer network) over 100 time steps. I'd like to just make sure that we are not sacrificing too much speed with this. No rush as I will not get to merging this in for a little bit... Thanks!

chad-klumb:
Timing for this branch should be identical to #558, so < 5% difference for the usual EpiModelHIV example.

There are basically no changes to the code in this PR, just moving what was left of tergmLite into EpiModel.

AdrienLeGuillou:
So I ran most models in the Gallery and in NME with `EpiModel@master` and `EpiModel@post_tergmlite`. Overall the performance were very similar. Almost no difference in memory usage and from -5% to +15% runtime. 
I used profvis to compare the runs with both setup.

PR #564, opened 2021-07-21
fixes issue 563 (set_transmat function bug)

### LINKED ISSUE #563: set_transmat function
Hello,

If I understood the code correctly, I noticed that maybe it would be better to change this line code here:
https://github.com/statnet/EpiModel/blob/e402f7a427c9170108f714c10709efe5d2fe2686/R/net.mod.infection.R#L376

to something like:

```
if(!is.null(dat$stats$transmat) == TRUE){
    del <- rbind(dat$stats$transmat, del)
}
```

that is because the first transmission may not happen at time = 2, for example. So https://github.com/statnet/EpiModel/blob/e402f7a427c9170108f714c10709efe5d2fe2686/R/net.mod.infection.R#L376
will always evaluate to true. 

And it will do the same "command" twice:
https://github.com/statnet/EpiModel/blob/e402f7a427c9170108f714c10709efe5d2fe2686/R/net.mod.infection.R#L377
and
https://github.com/statnet/EpiModel/blob/e402f7a427c9170108f714c10709efe5d2fe2686/R/net.mod.infection.R#L380

Cheers,
Fabricia.

AdrienLeGuillou:
fix #563

### DIFF
--- a/R/net.mod.infection.R
+++ b/R/net.mod.infection.R
@@ -373,11 +373,8 @@ set_transmat <- function(dat, del, at) {
   del[["sus"]] <- get_unique_ids(dat, del[["sus"]])
   del[["inf"]] <- get_unique_ids(dat, del[["inf"]])
 
-  if (at != 2) {
-    del <- rbind(dat$stats$transmat, del)
-  }
-
-  dat$stats$transmat <- del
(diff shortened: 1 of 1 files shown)


AdrienLeGuillou:
right now the test are still failing because `ndtv` is not available on CRAN anymore

smjenness (approved):

Issue #554, opened 2021-06-25
Create an extraction method for param.net class post netsim


smjenness:
Currently in a model with random parameters, there is not an easy way to extract the random parameters across simulations. One must look up `sim$param$random$param.values`. I think it would be helpful to have a function that would be run on a netsim class object that would look into sim$param and output a data frame of both fixed and random parameters. For example, the output would look like this:

```r
sim    rec.rate    act.rate    inf.prob
1      0.02        2           0.1434
2      0.02        1           0.1799
3      0.02        3           0.2400
```

where `act.rate` and `inf.prob` are random and `rec.rate` is fixed.

In one sense as.data.frame.print.netsim could do this, but I don't want people reaching into a netsim object with $. Instead, perhaps, get_param_list could do this without an output type argument?

AdrienLeGuillou:
I really like the `data.frame` way of showing the parameters, unfortunately, because the parameters are not necessarily scalars, it would mean that sometimes the columns would be list columns:

```
sim    rec.rate    act.rate    inf.prob    stratified.param
1      0.02        2           0.1434           <list>
2      0.02        1           0.1799           <list>           
3      0.02        3           0.2400           <list>
```

what I could do is have the function output a list of size `nsims` with each element being a "param.net" list with all parameters considered fix. The function could also have an argument to get only the parameters from some of the simulations and not all of them.

``` r
sim <- netsim(est, param, init, control)

get_param_list(sim, simno = c(1, 4, 12) # return a list of 3 "param.net" objects
```

What do you think @smjenness?

smjenness:
Hmmm... 

If the stratified parameters will almost always be a numerical vector, maybe we can just expand it into numbered columns like this

```r
sim    rec.rate    act.rate    inf.prob    stratified.param.1   stratified.param.2
1      0.02        2           0.1434        0.53                     0.78
2      0.02        1           0.1799        0.39                    0.43
3      0.02        3           0.2400        0.89                  0.38
```

AdrienLeGuillou:
On it

AdrienLeGuillou:
The PR above creates a function `get_param_set` that outputs:

```
sim    rec.rate    act.rate    inf.prob    stratified.param_1   stratified.param_2
1      0.02        2           0.1434        0.53                     0.78
2      0.02        1           0.1799        0.39                    0.43
3      0.02        3           0.2400        0.89                  0.38
```

as always, I am not great at naming functions so I can change it for a better one

[...]

PR #565, opened 2021-07-21
extract parameters

AdrienLeGuillou:
addresses #554


R/get.R
```diff
@@ -508,7 +508,7 @@ get_args <- function(formal.args, dot.args) {
 #' get_param_set(mod)
 #' @export
 get_param_set <- function(sims) {
-  if (class(sims) != "netsim") {
+  if (!any(class(sims) %in% "netsim")) {
```
smjenness, line 511: another function for this is: `inherits(sims, "netsim")`

AdrienLeGuillou: I understood where was the problem with the object having multiple classes. 
With a `netsim` object it works perfectly as it is only a "netsim". However, in the tests I try using `get_param_set` on a "control" object. I get the error as expected but also a warning because the `control` object is of class `c("control.net", "control.list", "list")`and the `if` statement only checks the first one. 

Using `inherits` as you suggest will solve that and make the code work as expected each time

smjenness (approved):

PR #575, opened 2021-08-25
create dat a minimal

### LINKED ISSUE #567: make a `create_dat_object` function to initialize a minimal `dat` object
we should have a small helper function `create_dat_object` that initialize it with all the required elements and the right classes. 

It can be used at the begining of a model simulation or in examples that require a functioning `dat` object.

Once it is made, all examples that work on `dat` should be updated with it in order to remove the `dontrun` statement.

AdrienLeGuillou:
Fix #567 

This PR creates a helper function to initialize a minimal `dat` object.

This first implementation is extremely simple and only offers to pass `param`, `init` and `control` to populate the `dat` object. There is an option to pass nothing which results in the creation of simple shell list with all the named elements as empty lists. This option will be useful for testing purposes as the new APIs generally work only on a small subset of the `dat` object.

I did not included options for setting up the networks for now as the differences there is between EpiModel and EpiModelHIV require either a function that handles all cases or the generalization of one way. I feel the second option is more viable long term and this can be added later on to the `create_dat_object` function.


AdrienLeGuillou:
The current implementation is very simple but can be made more complex as we implement new APIs that relies on `dat` to have a given structure.

I also implemented a `_timestep` element in this newly created object. It is set to 1 by default and can be manipulated by 
- `get_current_timestep`: to get `at` even with just `dat`
- `increment_timestep`: that would add 1 to the inner `_timestep` value
- `set_current_timestep`: that sets it to any value

This solves #368. I named the element "_timestep" to prevent a user to unintentionally changing it's value. For now the value is set to be the `at` value in `netsim_loop`. The rest of the implementation remains unchanged. Having this in will allow some accessors to work without the need to pass `at` as an argument. I feel that `at` should be passed as a argument only when we want the user to be able to get/set a value at a time other than the current timestep. For instance in PR #579 relative to cumulative edgelist, `update_cumulative_edgelist` needs `at` to calculate the age of a relationship so it should be pulled directly from `dat`.

PR #579, opened 2021-09-07
Fix576 cumulative el

### LINKED ISSUE #576: cumulative partnership list
We need a general way to build a cumulative partnership list

requirements:

- allow truncation (remove partnership that ended more than `N` time step ago)
- use `unique_ids` to handle the death
- should work with one or many networks (see `EpiModel::discord_edgelist`)
- should be flexible in how the list can be filtered

terminology:

- “cumulative edgelist”
- could potentially be stored on dat as `dat$el_cuml`, so it is similar but distinct from dat$el

AdrienLeGuillou:
Fixes #576 

The current implementation adds the following helper functions:

- `get_edgelist(dat, at, network)` returns a matrix edgelist for the specified network using the appropriate function depending on the network types
-  `get_cumulative_edgelist(dat, network)` returns the current cumulative edgelist for the specified network
-  `update_cumulative_edgelist(dat, at, network)` update the cumulative edgelist for the specified network
- `get_cumulative_edgelists_df(dat, networks = NULL)` returns the cumulative edgelists of the specified networks (all by default) in one `data.frame` with a "network" column specifying in to which network each edge belongs

in addition, an optional `cumulative_edgelist(dat, at)` module is implemented. It calculates the cumulative edgelists for all the networks. It uses the "truncate.el_cuml" control to truncate the cumulative edgelists for old edges. (this control is used inside the `update_cumulative_edgelist` function. 

The implementation works with and without `tergmLite == TRUE`

### DIFF
--- a/R/el_cuml.R
+++ b/R/el_cuml.R
@@ -13,33 +13,42 @@ get_cumulative_edgelist <- function(dat, network) {
 
   if(is.null(el_cuml)) {
     el_cuml <- data.frame(
-      head = numeric(0),
-      tail = numeric(0),
+      head  = numeric(0),
+      tail  = numeric(0),
       start = numeric(0),
-      stop = numeric(0)
+      stop  = numeric(0)
     )
   }
 
   return(el_cuml)
 }
 
-update_cumulative_edgelist <- function(dat, network) {
+update_cumulative_edgelist <- function(dat, at, network) {
   truncate.el_cuml <- get_control(
-    dat, "truncate.el_cuml", override.null.error = FALSE)
-  truncate.el_cuml <- if (is.na(truncate.el_cuml)) Inf else truncate.el_cuml
+    dat, "truncate.el_cuml", override.null.error = TRUE)
+  truncate.el_cuml <- if (is.null(truncate.el_cuml)) Inf else truncate.el_cuml
 
   el <- get_edgelist(dat, network)
-  el_cuml <- get_edgelist(dat, network)
+  el_cuml <- get_cumulative_edgelist(dat, network)
 
   el <- data.frame(
     head = get_unique_ids(dat, el[, 1]),
     tail = get_unique_ids(dat, el[, 2]),
-    current = T
+    current = TRUE
   )
 
-  el_cuml <- merge(el_cuml, el, by = c("head", "tail"), all = T)
-  el_cuml[is.na(el_cuml[["start"]]), ]$start <- at
-  el_cuml[is.na(el_cuml[["current"]]), ]$stop <- at
+  el_cuml <- merge(el_cuml, el, by = c("head", "tail"), all = TRUE)
+
+  new_edges <- is.na(el_cuml[["start"]])
+  if (any(new_edges)) {
+    el_cuml[new_edges, ][["start"]] <- at
+  }
+
+  if (at == 90) browser()
+  terminated_edges <- is.na(el_cuml[["current"]]) & is.na(el_cuml[["stop"]])
+  if (any(terminated_edges)) {
+    el_cuml[terminated_edges, ][["stop"]] <- at - 1
+  }
 
   if (truncate.el_cuml != Inf) {
     rel.age <- at - el_cuml[["stop"]]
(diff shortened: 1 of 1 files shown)


R/el_cuml.R
```diff
@@ -0,0 +1,132 @@
...
+#' @param dat a Master list object of network models
+#' @param at the current timestep
+#' @param network the index of the network to extract the edgelist from
+#'
+#' @return an edgelist in matrix form with the two columns. Each column
+#'          contains the posit_ids (see \code{get_posit_ids})of the nodes in
+#'          each edge
+#'
+#' @export
+get_edgelist <- function(dat, at, network) {
```
smjenness, line 16: this is helpful, but do you then integrate it in `discord_edgelist`?

R/el_cuml.R
```diff
@@ -0,0 +1,132 @@
...
+#' To avoid storing a cumulative edgelist to long, the "truncate.el_cuml"
+#' control value defines a number of steps after which an edge that is no longer
+#' active is truncated out of the cummulative edgelist.
+#'
+#' @return an updated Master list object of network models
+#'
+#' @export
+update_cumulative_edgelist <- function(dat, at, network) {
+  truncate.el_cuml <- get_control(
+    dat, "truncate.el_cuml", override.null.error = TRUE)
```
smjenness, line 70: I prefer consistent dot case for arguments. So `truncate.el.cuml`

R/el_cuml.R
```diff
@@ -0,0 +1,132 @@
...
+#'                 edgelists from. If NULL (default) extract all the cumulative
+#'                 edgelists.
+#'
+#' @return a \code{data.frame} with 5 columns:
+#'          head and tail, the unique_ids (see \code{get_unique_ids}) of the
+#'          nodes on the edge. Start and stop, the timestep where the edges
+#'          started and stopped. network, the network on which the edge is.
+#'
+#' @export
+get_cumulative_edgelists_df <- function(dat, networks = NULL) {
```
smjenness, line 118: is this just for post processing of the netsim object? Just curious where this is used.

R/net.mod.cumulative_edge_list.R
```diff
@@ -0,0 +1,37 @@
...
+#' be accessed using \code{\link{get_cumulative_edgelist}}
+#' \code{\link{get_cumulative_edgelists_df}}
+#
+#' @seealso \code{\link{get_cumulative_edgelist}}
+#' \code{\link{get_cumulative_edgelists_df}} \code{\link{netsim}}
+#'
+#' @export
+#' @keywords netMod internal
+#'
+cumulative_edgelist.net <- function(dat, at) {
```
smjenness, line 31: I don't think this should be a standalone, top-level module. It is just two functions (or sets of functions) to update and then get the cuml edgelist that would be called within two places: nwupdate.net (for the updating) and any module that needs that cuml el data (e.g., a contact tracing module).

smjenness: Because of that, I would suggest setting the control.net setting `truncate.el.cuml` default to 0, which means, don't store this data at all. If `truncate.el.cuml >= 1`, then it gets run. It doesn't need to get saved to the output netsim object, but it could potentially if we specified `el_cuml` in `save.other` in `control.net`

smjenness (changes requested):
See comments above

smjenness:
In addition to comments above, I feel like it would be helpful to have a simple "contract tracing" type of function that would operate on the `el_cuml` object --> given a vector of index IDs (e.g., of infected persons), return a list of contact/partner IDs for each person. Maybe this is a nested list? There should also be a argument to this to truncate to only active (alive) contacts. What do you think?

R/el_cuml.R
AdrienLeGuillou: I think I should! I will make sure that it works similarly to [this block](https://github.com/statnet/EpiModel/blob/fix576_cumulative_el/R/net.mod.infection.R#L310-L322) in `discord_edgelist` (I think I missed one possibility in my function) and then use it in `discord_edgelist` as well

R/el_cuml.R
AdrienLeGuillou: Right now it is not used. But I was thinking of what we do in CombPrevNet where this is the format we would want to work with. In general I feel that having things stored individually inside a `list` gives the best flexibility but combined `data.frame`s are better suited for analysis and filtering for custom module. As it can be tricky to consistently combine `data.frame`s into one without errors (especially in corner cases, where one is empty for instance) I thought it would be useful to have the helper function directly in so we can test it thoroughly and module builders will be able to use it without risking unforeseen errors.

R/net.mod.cumulative_edge_list.R
AdrienLeGuillou: I agree with all this. I will make the changes

AdrienLeGuillou:
agreed for the contact tracing function.

maybe something like this:

`get_partners(dat, index_posit_ids, networks = NULL, max.age = NULL, only.active = FALSE)`
- `dat`: the master object
- `index_posit_ids`: should we ask for unique_ids instead? I think using posit_ids makes more sense 
- `networks`: on which networks to look? (similar to `get_cumulative_edgelist_df`)
- `max.age`: should we only consider partnerships that ended less than `max.age` steps ago?
- `only.active`: should we filter out inactive contacts?

return: either a nested `list` or a `data.frame` similar to the one in `get_cumulative_edgelist_df` but with columns:
- `index`: so the index is always in first position
- `partner`: the partners are always in second position
- `start`
- `stop`
- `network`

the `data.frame` may be easier to work with compared to the nested list that will be pretty ragged and will require either a good knowledge of the `list` functions to work with (`Filter`, `Reduce`, ...) or a conversion to `data.frame` 

Usual disclaimer for the quality of the names

R/net.mod.cumulative_edge_list.R
AdrienLeGuillou: New thoughts on that while making it:

I think we might want another control like `cumulative.edgelist` (logical) to ask whether or not to compute the cumulative edgelist and then `truncate.el.cuml`. 

The reasoning is that by default we probably don't want to calculate the `el_cuml` but activating that with the truncation control feels weird. Also, sometimes we might want to calculate the `el_cuml` with `truncate.el.cuml == 0` to keep only the active partnerships but still know their start step. 

What do you think? 

Also, if we go this route, if `truncate.el.cuml` is `NULL`, should we consider that it's `0` or `Inf` (current behavior)

PR #581, opened 2021-09-16
fix #580 plot.netsim support for cex.lab and cex.axis

### LINKED ISSUE #580: plot.netsim support for cex.lab and cex.axis
As far as I can tell, `plot.netsim` doesn't support calling a number of basic but important parameters such as `cex.lab` and `cex.axis`. I'm not sure if this is by design (the built-in plotting meant mainly for quick visualizations) but may be worth adding some of these features.

Of course one can also work around some of this by setting `xlab=''` and then using `mtext` with side =1.

AdrienLeGuillou:
#580

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -2430,46 +2430,40 @@ plot.netsim <- function(x, type = "epi", y, popfrac = FALSE, sim.lines = FALSE,
 
     ## Missing args ##
     if (is.null(da$xlim)) {
-      xlim <- c(0, nsteps)
-    } else {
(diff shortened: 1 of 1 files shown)


AdrienLeGuillou:
This new implementation of the plot functions for `type = "epi"` and `type = "formation"` sets the default arguments inside the `da` (dot arguments) list if they are not provided by the user then call the `plot` function with `do.call(plot, da)`.

This allows the user to pass arbitrary arguments to the `plot` functions without loosing the default provided by `plot.netsim`

PR #592, opened 2021-10-20
Add support for heterogeneous dissolution model diagnostics

### LINKED ISSUE #508: Support heterogeneous dissolution models to print.netdx and plot.netdx
specifically nodematch and nodemix

sgoodreau:
Closes #508

### DIFF
--- a/R/netdx.R
+++ b/R/netdx.R
@@ -104,6 +104,20 @@
 #' plot(dx2, type = "duration")
 #' plot(dx2, type = "dissolution", qnts.col = "orange2")
 #' plot(dx2, type = "dissolution", method = "b", col = "bisque")
+#' 
+#' # Dynamic diagnostics on a more complex model
+#' nw <- network_initialize(n = 1000)
+#' nw <- set_vertex_attribute(nw, 'neighborhood', rep(1:10,100))
+#' formation <- ~edges+nodematch('neighborhood', diff=TRUE)
+#' target.stats <- c(800,45,81,24,16,32,19,42,21,24,31)
+#' coef.diss <- dissolution_coefs(dissolution = 
+#'                      ~offset(edges)+offset(nodematch('neighborhood',diff=TRUE)),
+#'                      duration = c(52,58,61,55,81,62,52,64,52,68,58))
+#' est <- netest(nw, formation, target.stats, coef.diss, verbose = FALSE)
+#' dx11 <- netdx(est, nsims = 5, nsteps = 100)
+#' plot(dx11)
+#' plot(dx11, type = "duration", plots.joined=TRUE, qnts=0.2)
+#' plot(dx11, type = "dissolution", mean.smooth=FALSE, mean.col='red')
 #' }
 #'
 netdx <- function(x, nsims = 1, dynamic = TRUE, nsteps,
--- a/man/netdx.Rd
+++ b/man/netdx.Rd
@@ -139,6 +139,20 @@ plot(dx2, stats = c("edges", "meandeg"), plots.joined = FALSE)
 plot(dx2, type = "duration")
 plot(dx2, type = "dissolution", qnts.col = "orange2")
 plot(dx2, type = "dissolution", method = "b", col = "bisque")
+
+# Dynamic diagnostics on a more complex model
+nw <- network_initialize(n = 1000)
+nw <- set_vertex_attribute(nw, 'neighborhood', rep(1:10,100))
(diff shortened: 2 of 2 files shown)


sgoodreau:
@smjenness Looks like when I compile EpiModel now I get the following additional text in `src/RcppExports.cpp`. A quick web search suggests that this is normal for the latest version of the `Rcpp` package.  OK if I commit and push this? If so then I'm guessing you all may need to update `Rcpp` too to prevent the file from changing back and forth.

```
#ifdef RCPP_USE_GLOBAL_ROSTREAM
Rcpp::Rostream<true>&  Rcpp::Rcout = Rcpp::Rcpp_cout_get();
Rcpp::Rostream<false>& Rcpp::Rcerr = Rcpp::Rcpp_cerr_get();
#endif
```

sgoodreau:
@smjenness @AdrienLeGuillou 

Currently, `edgelist_meanage` is an exported function. I'm not sure whether anyone ever uses it directly, rather than via the call stack `netdx -> make_dissolution_stats -> edgelist_meanage`. (Note that `make_dissolution_stats` is an internal function only, and seems to be the only place that calls `edgelist_meange` in the code base, and is itself only ever called from `netdx`.

Either way, `edgelist_meanage` is now going to need more inputs than it used to, specifically the dissolution formula in some form or other.  If we knew that the function was only ever called down from `netdx` itself, then we don't have to repeat all of the processing of the dissolution formula into its constituent parts, and subsequent error checking, that occurs in `netdx` and/or `make_dissolution_stats`. If it can be called directly, then we need to repeat all of that.  We can't just move it from the higher functions down into this one because it's needed for other reasons up there too.

Should we repeat all that code (probably 30-40 lines)? Or should we stop exporting it, thus breaking backwards compatibility if there are users who call it directly?  Or is there a more elegant solution altogether?

sgoodreau:
> Should we repeat all that code (probably 30-40 lines)? Or should we stop exporting it, thus breaking backwards compatibility if there are users who call it directly? Or is there a more elegant solution altogether?

Note that to avoid duplicating all the code, we'd also need to remove the option of passing a `netest` object directly, rather than as an edgelist and now dissolution term info, since the former case would still require the processing. Note that only the edgelist option is used internally in EpiModel.

AdrienLeGuillou:
@sgoodreau I looked at the project I worked on (multiple EpiModelHIV and EpiModelCOVID), on none of them `edgelist_meange` was used. So as far as I know it is safe to remove from the export.

smjenness:
@sgoodreau : I agree with Adrien, you can remove that as an export. Note that I just merged main into your branch. There were a bunch of mostly cosmetic updates across EpiModel that your branch should be updated with. This shouldn't impact your work.

sgoodreau:
@smjenness @AdrienLeGuillou 

Thanks. Pulling in main was going to be my next step after getting my permissions working again, so thanks for taking care of it. And it allowed me to practice my stashing.

I also notice that both `edgelist_meanage` (exported) and `make_dissolution_stats` (not exported) currently have help pages, the former fleshed out and the latter very minimal.  Is it normal/desirable to have help pages for non-exported functions? If so, I can update the `edgelist_meanage` one even after removing the export. Or I can delete it, and perhaps `make_dissolution_stats` too while we're at it.

AdrienLeGuillou:
I would say that the more documentation the better, even for internal functions. When making such functions I try to at least put a minimal ROxygen header at least to describe the arguments and the output

smjenness:
@sgoodreau : at first glance, this looks great! It looks like you have one unit test added for all this functionality. Should we be testing against multiple dissolution model parameterizations? different terms in the model, different types of attributes involved?

@AdrienLeGuillou : could you also review this work. You might also have some suggestions about unit testing as you go.

smjenness:
Also, should we consider moving a packaged test to an example within netdx or plot.netdx?

sgoodreau:
@smjenness thanks.  I'm happy to add more tests, and examples. I didn't have a good sense of the volume of these you wanted and had thought to ask you about this earlier, but then the weird testing issues with the string parsing sidetracked me.  You want me to write them, or someone else? Now, or after Chad finalizes that additional `netdx` work?

smjenness:
@sgoodreau : perhaps for now, just a nodemix test in the mix too. The goal for the unit tests in general would be predict a set of conditions that could break the code. If you have other ideas of branching conditions for those types of use cases, please feel free to add those too.

sgoodreau:
OK tests and example added

smjenness (approved):

Issue #594, opened 2021-10-29
More informative attr list misalignment error message


smjenness:
Currently `deleteAttr` outputs this:

```r
  if (length(unique(sapply(attrList, length))) != 1) {
    stop("attrList must be rectangular (same number of obs per element)")
  }
```

which is triggered typically when someone forgets to update new attributes for arrivals/births.

It would be more helpful to have a more informative error message here. That could include to check the arrivals module, and a more specific listing of the attribute(s) in question. For example, if the length of the active attribute is the gold standard, then the length of any attributes not equal to active could be output in a little table.

smjenness:
Closed with #597

PR #597, opened 2021-11-04
better message in deleteAttr

AdrienLeGuillou:
related to #594 

With this PR `deleteAttr` will compare the length of every attribute to the length of the `active` attribute, created by the `append_core_attribute` function. 

If some attribute do not have the right length, an error detailing which attribute is faulty will occur.


R/utils.R
```diff
@@ -91,8 +91,17 @@ deleteAttr <- function(attrList, ids) {
...
-  if (length(unique(sapply(attrList, length))) != 1) {
-    stop("attrList must be rectangular (same number of obs per element)")
+
+  attr_length <- vapply(attrList, length, numeric(1))
+  expected_length <- length(attrList[["active"]])
+  wrong_length_attr <- names(attr_length)[attr_length != expected_length]
+
+  if (length(wrong_length_attr > 0)) {
+    stop(
+      "The following attribute do not have the right number of elements: \n",
```
smjenness, line 101:
```suggestion
      "The following attributes do not have the correct number of elements: \n",
```

R/utils.R
```diff
@@ -91,8 +91,17 @@ deleteAttr <- function(attrList, ids) {
...
+      paste0(wrong_length_attr, collapse = ", "),
+      "\n\n", "Check if they are initiated when new nodes are created."
```
smjenness, line 103:
```suggestion
      "\n\n", "Check if they are initialized when new nodes are created."
```

PR #613, opened 2021-11-29
Extend `updater.net` to controls

### LINKED ISSUE #606: Add alternative parameterization for resimulation.networks to allow for ranges
This has come up in the new CEA example in the EpiModel Gallery by @gknowlt. We need to stop the network resimulation at a specific time step to allow for other dynamic processes to accrue in the absence of ongoing ID exposure. I am thinking this would involve inputting a vector of time steps to control.net's resimulate.network, where the current approach is just TRUE or FALSE. If the vector < nsteps in control.net, then the resimulate.net module would get skipped. I'll ask @AdrienLeGuillou to work on this and @gknowlt to check it against the gallery example when done.

AdrienLeGuillou:
This PR generalise the `updater.net` module to allow it to modify
controls as well as parameters during runtime.

This can be used for instance to fix #606 by switching
`resimulate.network` on and off at specific timesteps


smjenness (changes requested):
Thanks for getting this done so quick. One requested change is to modify the unit test or add a new one that specifically tests updating resimulate.networks. Is there a logical expectation that you could specifically test against there? e.g., extract the networkDynamic object and see how many time slices there are? It should be < the nsteps.

smjenness (approved):

PR #614, opened 2021-12-08
Mention control_updater in NEWS

AdrienLeGuillou:
Should I also modify the "Working with model parameters" vignette to add a section covering the control updaters?

Maybe "## (Advanced) Time-Varying Controls" with a succinct example as the mecanisms are the same as for Time-Varying Parameters

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+## EpiModel 2.2.1
+
+### NEW FEATURES
+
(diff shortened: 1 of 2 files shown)


smjenness:
Maybe just a brief mention in the vignette. I don't think it will be as used by external users as the time-varying parameters.

smjenness:
You put the update about the control updater under EpiModel 2.2.0. You need to start a new version header and put the update there instead.

AdrienLeGuillou:
Oh! I forgot that 2.2.0 was already released

smjenness (approved):

PR #618, opened 2021-12-14
Change defaults of cuml el, add explicit args

smjenness:
This changes the default of cumulative edgelist storage to FALSE, and the default truncation to 0 (current edges only). Additionally, I added cumulative.edgelist and truncate.el.cuml as explicit args in control.net. 

@AdrienLeGuillou Could you check this PR?

### DIFF
--- a/R/net.mod.nwupdate.R
+++ b/R/net.mod.nwupdate.R
@@ -2,7 +2,7 @@
 #' @title Dynamic Network Updates
 #'
 #' @description This function handles all calls to the network object contained
-#'              on the master dat object handled in \code{netsim}..
+#'              on the master dat object handled in \code{netsim}.
 #'
 #' @param dat Master list object containing a full \code{networkDynamic} object
 #'        or networkLite edgelist (if using tergmLite), and other initialization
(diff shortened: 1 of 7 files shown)


R/net.mod.nwupdate.R
```diff
@@ -109,11 +109,11 @@ nwupdate.net <- function(dat, at) {
   }
 
   # Cummulative edgelist
-  if (!is.null(cumulative.edgelist) && cumulative.edgelist) {
+  if (!is.null(cumulative.edgelist) && cumulative.edgelist == TRUE) {
```
AdrienLeGuillou, line 112: not sure why you added the `== TRUE` 
it is usually considered a bad practice to compare a "truth value" to TRUE or FALSE and instead it's recommended to prefer the use of the expression itself, potentially with `!` for negation. (There are tons of jokes about that on r/ProgrammerHumor)
(Otherwise to be consistent we should use `is.null(cumulative.edgelist) == FALSE && cumulative.edgelist == TRUE`)

Even though adding the test `cumulative.edgelist == TRUE` does not add a lot of overhead, it can lead the reader to believe that the `== TRUE` does some conversion / specific R trick to make the code work. I personally always wonder when I see such a construct if there is an edgecase being solved by this but I am pretty sure this particular construct cannot solve anything as `if ()` already does that internally.

Overall, not a huge problem for me, but I would like to hear why you prefer this construct.

smjenness: I suppose just out of convention (something I have always done), to demonstrate that the object is logical. I realize it's not necessary. I'm fine with removing here, but then we should go through the entire EpiModel code base to make this consistent.

AdrienLeGuillou: Again, that would not change a lot in terms of speed so it's not an "issue". However, I am pretty sure I used the other convention everywhere in my code. So one way or the other there will be a need to harmonize :)
On that note, I think having a "style / coding guide" for EpiModel could be usefull. I will create an abstract ToDo on the project board

smjenness: Sounds good. But with respect to the substantive task at hand (this PR), are the default changes implemented here looking ok?

AdrienLeGuillou: Oh sorry, yes! They are perfect and definitively better defaults!

AdrienLeGuillou (approved):

PR #625, opened 2022-01-26
Utilities to export and load model scenarios from `data.frames`

AdrienLeGuillou:
This PR will provide utilities to apply scenarios to an EpiModel simulation.
A scenarios is a set of parameters to be changed into a pre-calibrated model

- transform a `list` of parameters into a specially formatted `data.frame`
- transform back these `data.frames` into `list` of parameters
- apply a scenario to a pre-calibrated model, either directly or at a specific time step during the simulation


AdrienLeGuillou:
list of changes:

- export 2 functions `use_scenario` and `make_scenarios_list`
- remove `updater.net` from the export as the functionality is now a default on `netsim`. The function itself as been renamed `input_updater` and is called internally in `netsim_loop`. (the change of name is to avoid confusion with other modules)
- rename `param.updater.list` and `control.updater.list` to `.param.updater.list` and `.control.updater.list`
- The vignette "working with model parameters" has been updated:
    - first section about the scenarios
    - second about the random parameters as before
    - two section with the (advanced) mention describing the use of the *updaters* functionalities outside of the scenarios. It has be updated to reflect the changes, most notably the fact that it now works with all network models, not only the `type = NULL` ones

AdrienLeGuillou:
@smjenness 
I made the function to create a `param` object from a `data.frame` containing 3 columns, "name", "value" and "type". Any other column is allowed as long as these 3 are there. 
This way we can have a descriptive excel sheet with sources and details and use that to make the `param` object.
I called the function `param.net_from_table` for now but am not really happy with it. If you have a suggestion for the name, I am all ears.

smjenness:
@AdrienLeGuillou : everything looks great with the underlying functionality. I made a bunch of edits to the parameters vignette at [7a0fb77](https://github.com/EpiModel/EpiModel/pull/625/commits/7a0fb7729a44682b2e7352b41597e842aeadc0ae). Could you just take a quick review? Once that is done, we will be ready to merge this.

AdrienLeGuillou (commented):
@smjenness 
Once again your changes are making things much clearer!
(I cannot "approve" the PR as I am the initial author)

Issue #641, opened 2022-03-03
Allow param.net to receive a formatted `data.frame` of parameters


AdrienLeGuillou:
add a new parameter to `param.net` to use a formatted `data.frame` of parameters.

The `data.frame` will need to contain at least the 3 following columns:
1. 'param': the name of the parameter. If it is a multi dimension parameter, prepend the name, with "_1", "_2", etc for the position in the vector.
2. 'value': the value for the parameter (or the Nth position if  multidimensional) 
4. 'type': a character string containing either "numeric", "logical" or "character". The type the value will be cast to.

If a parameter is defined twice (in the `data.frame` and as a normal argument), a WARNING will be produced and the value actually used will be displayed for the user. @smjenness, which value should be used? The one from the DF or from the argument to `param.net`

smjenness:
Closed with #642

PR #642, opened 2022-03-03
Param.net data.frame argument

AdrienLeGuillou:
addresses #641


AdrienLeGuillou:
I added the handling of a hidden parameters (in the ellipsis) called `data.frame.params`. It allows the setting of a bunch of parameters at once using the format we discussed. 
This element is handled before the formal argument. This means that if a parameter is defined in the data frame and in the call itself, the value from the call will be used.
I added a `@section` in the documentation explaining how it works.

@smjenness We probably want to have a section in the parameters vignette for it as well, I'll leave this PR as a draft until I finished it and tested the code with EpiModelHIV.

Finally, I left the formal arguments as is (i.e. I did not add `arg = NULL` to the function signature) to keep it consistent with `control.net` and `init.net`. Making such a change would only make sense if performed globally and is not necessary for this PR.

smjenness:
@AdrienLeGuillou : I made a few further edits to the docs. Does everything look ok to you?

AdrienLeGuillou:
@smjenness : thanks for these modifications, they really makes it clearer. Everything looks good to me.

PR #644, opened 2022-03-21
trim_netest

### LINKED ISSUE #550: Is there any way to reduce the size of netest object with network attributes without compromising netsim?
When dealing with a large network and the need for multiple `edgecov` variables, resulting `netest` objects end up taking up extraordinary amounts of memory because the underlying network is stored at least 4 times (`network`, `newnetworks`, etc.). For reasons I don't quite understand, large `netest` objects when used in `netsim` use more extra memory during simulation than simply the additive amount stored in the `netest` object itself (seems to be ~3-4x expansion). One potential solution is to store `edgecov` variables in the global environment rather than as network attributes using `%n%`, but this is not an option for me because of use in modules.

My question is whether there is any other way to potentially cut down on the size of the `netest` object to save memory. I've tried using sparse matrices, but estimation with `ergm` cannot allow this. Perhaps one or more of the network objects stored in `netest` could be modified/removed (or have their network attributes deleted)? Any other potential solutions?

chad-klumb:
closes #550

implements trim_netest to remove networks and environments from the netest object

also includes updates needed to use networkLites in netest/netdx/netsim


smjenness:
Hi @chad-klumb : thanks very much for this! Given the updates to the remotes fields in DESCRIPTION does this depend on pre-CRAN versions of statnet packages? If so, what are the expected timeframes for pushing those?

chad-klumb:
passing a `networkLite` as input to `simulate` and getting a `networkDynamic` as output requires statnet/tergm@55e390bd86e2d5bb57c5910d81bc6c8364db3663

as for time of CRAN push, I would ask @krivit

chad-klumb:
I added dropping of `fit$sample` and optional conversion to `networkLite` to the `trim_netest` function.  While `fit$sample` dropping and `networkLite` conversion can be controlled by the user prior to/in the `netest` call itself, it may be a bit simpler to just do it all at once afterwards.

smjenness:
@chad-klumb : could you take a look at this? I merged main into this branch to bring everything up to date, manually resolved some conflicts in tests/netdx in [faad4c9](https://github.com/EpiModel/EpiModel/pull/644/commits/faad4c920b1c2cfa295b7bdc827bbb105bdfd21f), but we are getting [this test failure](https://github.com/EpiModel/EpiModel/runs/5857537643?check_suite_focus=true).

Here is quick MRE that works on this branch. Is trim_netest not intended to work with netdx when used in parallel?

```r
num <- 50
nw <- network_initialize(n = num)
formation <- ~edges
target.stats <- 15
coef.diss <- dissolution_coefs(dissolution = ~offset(edges), duration = 20)
est1 <- netest(nw, formation, target.stats, coef.diss, verbose = FALSE)

est1 <- trim_netest(est1)

dx1 <- netdx(est1, nsims = 2, nsteps = 25, ncores = 2)
```

errors with:
```r
> dx1 <- netdx(est1, nsims = 2, nsteps = 25, ncores = 2)

Network Diagnostics
-----------------------
- Simulating 2 networks
 Error in { : 
task 1 failed - "no applicable method for '%ergmlhs%' applied to an object of class "NULL""
```

Thanks!

chad-klumb:
I will look into it

smjenness (approved):

Issue #573, opened 2021-08-19
updating EpiModel if separable wrappers are deprecated/removed from tergm


chad-klumb:
Some desire to deprecate/remove the stergm function (and presumably also the separable simulation wrappers and separable control functions) from the tergm package has been expressed.  Doing so would require updating EpiModel.

If these functions are removed from tergm, the lowest-impact option from the EpiModel user's point of view would probably be to keep EpiModel user interfaces as-is (in particular, separable) with the sole exception of replacing current separable controls with their non-separable equivalents, probably also changing the argument names from set.control.stergm to set.control.tergm for consistency.  This minimizes the need for users to change their code, but creates a bit of dissonance in that formulas, coefficients, etc. are specified separably while the control arguments are non-separable; I don't think that's a huge deal, but it is something to be aware of.  Beyond control arguments, the necessary conversions from separable to non-separable could be done internally in EpiModel, and are not things most users would need to concern themselves with.

A much higher-impact option would be to convert EpiModel to nonseparable interfaces.  Doing this properly would mean removing all reference to "formation" and "dissolution".  I don't see a great reason to convert EpiModel to nonseparable interfaces until such time as the additional generality provided by the non-separable interface has some meaningful application(s) in epidemic research, *and* we also have the ability to estimate such effects for research-level models.  In any event, this would require a much larger set of changes to EpiModel, many of which would be user-facing.

A third option is to leave separable functions in the tergm package as long as there is some reason to do so, recalling that maintaining backwards compatibility as much as possible was a top goal of the tergm 4.0 conversion process.  This would not require any changes to EpiModel (as long as we keep the separable functions in tergm).  The separable tergm functions probably represent fewer lines of actual code than the 3.9 term argument handling in ergm (which we've kept around for years and AFAIK have no immediate plans to remove).  The downside of this option is, obviously, more code in tergm.

chad-klumb:
Currently, in core EpiModel, the `stergm` function is called here:

https://github.com/EpiModel/EpiModel/blob/6955a42eadf3120e9a9b02e71ceaead50439afb3/R/netest.R#L169-L180

and separable simulation wrappers are called here:

https://github.com/EpiModel/EpiModel/blob/6955a42eadf3120e9a9b02e71ceaead50439afb3/R/net.mod.simnet.R#L52-L63

[...]

PR #645, opened 2022-03-22
deprecate separable interfaces in favor of nonseparable ones

chad-klumb:
references #573

when the time comes you may simply delete the separable calls, controls, documentation, tests, etc.


R/net.mod.simnet.R
```diff
@@ -48,23 +48,42 @@ sim_nets_t1 <- function(x, dat, nsteps) {
...
-                      monitor = nwstats.formula,
-                      nsim = 1,
-                      control = set.control.stergm)
-      sim <- networkDynamic::activate.vertices(sim, onset = 1, terminus = Inf)
-    })
+    set.control.tergm <- get_control(dat, "set.control.tergm")
+
+    if (!is.null(set.control.stergm)) {
+      # Simulate dynamic network
+      suppressWarnings({
```
AdrienLeGuillou, line 55: Why are we suppressing the warnings here? 
(pure curiosity as we were doing it prior to this PR as well)

AdrienLeGuillou (approved):

R/net.mod.simnet.R
chad-klumb: I think there are time stamp warnings in some of the `simulate` calls in `simnet`.  It may be possible to avoid them in typical use cases by adjusting `time.start` and `time.offset`.

AdrienLeGuillou: thanks for the explanation. I don't think it's a necessity to address that here though.

smjenness (approved):

PR #646, opened 2022-03-24
allow symbol search

### LINKED ISSUE #640: absdiffby in EpiModel
Dear all,

Since the term absdiffby has embeded included in `EpiModel`, I get an error during network fitting that was working when I was importing the term from `EpiModelHIV`.

Here a reproducible example

```
library(EpiModel)

n <- 1000
ntw <- network_initialize(n = n)
## add sex attribute 0 = female, 1 = male
set.vertex.attribute(ntw, "sex", sample(0:1, n, replace = TRUE))
## add age attribute
set.vertex.attribute(ntw, "age", sample(20:60, n, replace = TRUE))

deg.mean <- 0.6

fit <-
  netest(
    ntw,
    formation = ~ edges + absdiffby("age", "sex", 5)  + nodematch("sex", diff = FALSE),
    target.stats = c(n * deg.mean, n * deg.mean * 1,  0),
    coef.diss = dissolution_coefs(~ offset(edges), 20)
  )
```

which return the following error:

```
Error in summary.ergm_state(state) : 
  Error in ModelInitialize: term with functions EpiModel::absdiffby is declared to have statistics but does not appear to have a change or a difference function. Memory has not been deallocated, so restart R sometime soon.
```

Am I doing something wrong?

Thanks a lot,
Damien

Note:

Here is my `sessionInfo()`

```
> sessionInfo()
R version 4.0.5 (2021-03-31)
Platform: x86_64-redhat-linux-gnu (64-bit)
Running under: Fedora 34 (Workstation Edition)

Matrix products: default
BLAS/LAPACK: /usr/lib64/libflexiblas.so.3.1

locale:
 [1] LC_CTYPE=en_GB.UTF-8       LC_NUMERIC=C               LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8    
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8    LC_PAPER=en_GB.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] EpiModel_2.2.1        statnet.common_4.5.0  tergm_4.0.2           ergm_4.1.2            networkDynamic_0.11.0
[6] network_1.17.1        deSolve_1.30    
```

chad-klumb:
Fixes #640


smjenness:
@chad-klumb : Thanks for this, but isn't [RcppExports](https://github.com/EpiModel/EpiModel/blob/main/src/RcppExports.cpp) an auto generated file (by Rcpp) that we are not supposed to edit? I supposed it's ok for now until we write another C++ function with Rcpp (which is quite rare).

chad-klumb:
I did not edit RcppExports.cpp directly; I included `// [[Rcpp::init]]` in RegisterDynamicSymbol.cpp which causes `Rcpp::compileAttributes` to include a call to the corresponding function in its auto-generated RcppExports.cpp file.

smjenness:
Got it, thanks for clarifying

smjenness (approved):

Issue #549, opened 2021-06-17
Review and fix one step dissolution modeling


martinamorris:
Taking off from #525, it seems like there may be different options to consider for simulating these onetime partnership networks in `tergm`.  @krivit referenced [this issue in the tergm repo](https://github.com/statnet/tergm/issues/41) and pointed out that

````
simulate(nw ~ edges, dynamic=TRUE, time.slices=100, coef=theta, output="networkDynamic")
````
will generate a `networkDynamic` object each of whose 100 time steps is governed by whatever model/theta are specified.  Note this relies on the `tergm::simulate_formula.network()` function.

smjenness:
@ThomasKraft : No, you should be all set. All of the above is more related to consistent in network storage for simulations across different parameterizations. There was one issue related to time varying attributes that is indeed a bug (my comment directly above) but that only concerns ndtv network movies. Fingers crossed for your paper!

chad-klumb:
@smjenness Are you opposed to handling the "ergm" (duration = 1) case in the same way the "tergm" (duration > 1) case is currently handled, in the sense of how the networkDynamic is generated during the simulation, and using tergm controls (rather than ergm controls) to modify the simulation when needed?

chad-klumb:
(i.e., generating independent cross-sectional networks at successive time points when duration = 1, but recording e.g. edge activity in the networkDynamic following the conventions of the tergm package's simulate functions, and using control arguments for those functions)

chad-klumb:
implemented with tergm simulation in #648

if need be this can be changed to ergm simulation, but since netsim defines a time step, it is not entirely unreasonable to use tergm simulation even if the model is cross-sectional

Issue #528, opened 2021-05-20
Allow for static ERGM resimulation for duration = 1 models


smjenness:
Following up on #525. 

Need to allow for static ERGM resimulation here:
https://github.com/statnet/EpiModel/blob/master/R/net.mod.simnet.R#L87-L126

But also, redo network initialization, check network stats, and check reconstruction of networkDynamic object from panel ERGM sims.

ThomasKraft:
@smjenness Just trying to get a handle on when this implementation might be ready since I'll be using for the last step of an ongoing project -- am I correct that it should be a couple of months as part of Intermediate Issues? I'm not meaning to be pushy!

smjenness:
It may be a couple weeks

smjenness:
@ThomasKraft : sorry for the delay on this. It took quite a bit of work on the underlying infrastructure for network resimulation. This is now ready on #546, which will be merged into master shortly. 

Here's a working example:
```
# Your original working example of a very high density network
num = 20
toy <- network_initialize(num)

ft <- netest(toy,  formation = ~edges ,
             target.stats = 100,
             coef.diss = dissolution_coefs(dissolution = ~offset(edges),
                                           duration = 1),
             edapprox = TRUE)
param <- param.net(inf.prob = 0.2, act.rate = 1, rec.rate = 1/3)
init <- init.net(i.num = 10, r.num = 0)
control <- control.net(type = "SIR", nsteps = 25, nsims = 2, ncores = 1,
                       tergmLite = FALSE,  resimulate.network = FALSE, verbose = FALSE)
simz <- netsim(ft, param, init, control)

# get_network converts stored cross-sectional network list into a networkDynamic object
net <- get_network(simz, sim = 2)
net

# The timed edgelist now shows dyad spells of heterogeneous duration that span time steps 
nwd <- as.data.frame(net)
head(nwd, 25)

# You can also pull out the network list directly and work with that
nwl <- get_network(simz, sim = 2, ergm.create.nd = FALSE)

# network structure on first, second, third... time steps
nwl[[1]]
nwl[[2]]
nwl[[3]]
```

smjenness:
There are some more variations here, as I used these for unit testing: https://github.com/statnet/EpiModel/blob/resim-ergm/inst/test-static-ergm.R.

ThomasKraft:
[...]

Issue #525, opened 2021-05-11
within-step edge dissolution and formation (help Q)


ThomasKraft:
I would like to run a version of a model in which egocentric data are used to measure target stats but no data are available on the duration of ties. Instead of specifying `coef.diss = dissolution_coefs(dissolution = ~offset(edges), duration = X)`, I would like the duration of ties to be determined "naturally" based on the propensity of nodes with given attributes to interact on a repeated basis.  My thought on achieving this goal was to fit a model with `netest` that specifies a dissolution coefficient with `duration = 1`, such that a static ERGM is essentially re-estimated at each time point, e.g.:

```
 ft <- netest(toy,
                   formation = ~edges + various attributes,
                   target.stats = c(X, Y, Z),
                   coef.diss = dissolution_coefs(dissolution = ~offset(edges), duration = 1)
```

However, this approach fails of course in the following `netsim` (can't see the behavior using `netdx` because it won't do dynamic simulation for duration = 1) due to the separable property (independence within time-steps) of the STERGM formation/dissolution parts. If a tie 2-3 forms in step 4, then it necessarily dissolves in step 5 but cannot also re-form in step 5. This is of course expected behavior, but it thwarts my attempt to allow a distribution of tie durations that are governed by how often in a series of static ERGMs a dyad retains a tie based on existing attributes (and this pushes for greater mixing than might otherwise occur).

One potential option I am considering is to modify the `resim_nets` module such that all ties are automatically dissolved before resimulating the network, such that at each step all ties are eligible to form (as they were at initialization). But my hope is that someone here knows of a much less hacky solution! Any ideas? I can provide a reprex that demonstrates the behavior in the network edgelist if that were helpful.

smjenness:
@ThomasKraft : I think I understand the issue. If you use a duration of 1, then EpiModel interprets this as a duration-less network, i.e., a cross-sectional ERGM rather than a dynamic TERGM. The edgelist therefore exists within a time step and then gets refreshed the next time step. So a tie 2-3 that might exist within step 4, and could randomly exist in step 5 too. So no need to manually go in there to modify resim_nets. Does that clarify?

ThomasKraft:
@smjenness I've been puzzling over your comment given that all my tests have been showing different behavior than what you expected. After diving deeper, I'm now wondering if there is buggy behavior going on.

[...]

PR #648, opened 2022-03-29
updates to duration 1 handling

chad-klumb:
references #549

stacked on #645 as that introduces nonseparable interfaces which are required for the approach implemented here

the duration 1 case is handled by tergm simulation, producing a networkDynamic as in the duration > 1 case

nw_list has been removed


chad-klumb:
As noted in #549, there is a choice to be made as to whether we simulate the `duration = 1` case using `ergm` simulation functions and controls or `tergm` simulation functions and controls.  Either approach is possible and can be used to produce valid results.  For consistency of interface we should presumably use the same approach for both "full tergm" and "tergmLite" simulation within `EpiModel` (understanding that `networkDynamics` will be produced only in the "full tergm" case; "tergmLite" simulation will continue to operate based on `edgelists`/`networkLites`, as it does now).

Using `tergm` simulation functions and controls will result in somewhat cleaner `EpiModel` code because all the `networkDynamic` business is handled for us by the `tergm` package.

Using `ergm` simulation functions and controls would require us to handle conversions between `networks` and `networkDynamics` in `EpiModel`, possibly using some "internal" `tergm` functions (that are not currently exported, but could be) for updating the `networkDynamic` after the `ergm` simulation.  This will make `EpiModel` code in the "full tergm" case a bit longer/messier.

Upsides of using `ergm` simulation could include faster simulation times (probably most pertinent to the "tergmLite" case), and controlling the cross-sectional simulation using cross-sectional (i.e., `ergm`) controls.  (It's not clear that the runtime gain would be significant, but `ergm` simulation is probably at least a bit faster than `tergm` simulation for the same cross-sectional model.)  OTOH since `netsim` defines time steps, the simulation is arguably temporal even if the model is cross-sectional, and in that sense `tergm` controls are almost more appropriate.  But this is clearly a matter of opinion.

@smjenness Do you have a preference here?

smjenness:
@chad-klumb : thanks for your work on this. One of the considerations of ergm versus tergm for duration = 1 is whether the tergm approach can actually represent single time step edges. We explored that in #525 with changes implemented in #528. I found earlier that tergms did not allow for consecutive new edges between the same pairs. For example, if an edge existed between two nodes at time step 5, tergms represent that as spanning from steps 5 to 6, and therefore there could not be a new edge between the same two nodes until step 7. Although rare in many full-scale research models, it arose in one users case (the issue filer of #525) and I believe it conceptually makes more sense to have duration=1 models allow for consecutive repeat edges. That was the impetus for the changes in #528, and after I implemented that a bit of the core infrastructure around dynamic networks broke. I hope that addressed your questions, but let me know if not. Thanks!

chad-klumb:
`tergm` simulation can be used to generate the same dynamics as `ergm` simulation by simply using the `ergm` formula and coefficients with `dynamic = TRUE` in the `simulate` call

this differs from the EDA in that `Form` and `Diss` are not used

the EDA dynamics with duration 1 will lead to the non-consecutive behavior you described above

the ergm dynamics will not have this property; a dyad being an edge on consecutive time steps is perfectly allowed

you can see the difference in function calls as follows:

EDA dynamics (`Form` and `Diss` are used):

https://github.com/EpiModel/EpiModel/blob/3f5da0daec027aa92d161145313df655b75fbbe3/R/net.mod.simnet.R#L187-L196

`ergm` dynamics (despite using `tergm` functionality since `dynamic = TRUE`):

https://github.com/EpiModel/EpiModel/blob/3f5da0daec027aa92d161145313df655b75fbbe3/R/net.mod.simnet.R#L200-L209

chad-klumb:
You can see how edge activity is encoded in this scheme in the following example:

```
require(tergm)
set.seed(0)
nw <- network.initialize(10, dir = FALSE)
nwd <- simulate(nw ~ edges, coef = c(0), time.slices = 5, dynamic = TRUE)
nwd$mel[[8]]
```

which produces

```
> nwd$mel[[8]]
$inl
[1] 6

$outl
[1] 5

$atl
$atl$na
[1] FALSE

$atl$active
     [,1] [,2]
[1,]    1    3
[2,]    4    5
```

This means the dyad `5-6` was an edge at times 1, 2, and 4, but not at 3 or 5.

*Given* that the edge was toggled off at time 3, it cannot be toggled back on until time 4, but this does not preclude being an edge at consecutive time steps, as seen at times 1 and 2.

smjenness:
@chad-klumb : excellent example, thanks! It looks like the tergm approach for duration = 1 models is the way to go; I am not too concerned with the runtime issues you noted above as those will be minor. Any other questions before you implement this? Thanks!

chad-klumb:
No other questions come to mind at the moment, just note that tergmLite/EpiModelHIV applications using a cross-sectional model will have their control arguments changed from ergm controls to tergm controls.

I should have these changes done before our next meeting.

smjenness:
@AdrienLeGuillou Please note the change to default controls needed for EpiModelHIV/EpiModelCOVID

chad-klumb:
should be okay to merge now

in `netsim`, the `set.control.ergm` argument is now only used during initialization, and the documentation has been updated accordingly

I've also opened an EpiModelHIV PR with analogous changes

smjenness (approved):

PR #685, opened 2022-05-04
Rebuild networkLite docs generates several errors

smjenness:
@chad-klumb : I rebuilt (`devtools::document`) the documentation on the main branch, which re-generated the NAMESPACE. It looks like these docs were not rebuilt for one of your PRs (I suspect #665) that we merged in? This subsequently causes an error with R CMD Check that you will see on Github actions.


chad-klumb:
Looks like you don't have the master version of network installed.

smjenness:
Shouldn't Github actions pull that from the Remotes field in the DESCRIPTION?

smjenness:
Relatedly (or maybe not), it looks like Pavel published a new version of [statnet.common](https://cran.r-project.org/web/packages/statnet.common/index.html), so could that be removed from the Remotes field now on EpiModel's DESCRIPTION?

Issue #684, opened 2022-05-04
formula parsing/checking in dissolution model processing


chad-klumb:
Issues regarding formula handling in EpiModel came up in #508.  Three functions were mentioned:

- `netdx`
- `dissolution_coefs`
- `diss_check`

The problem in `netdx` should be resolvable by using `ergm.godfather` to determine dyad type in heterogeneous dissolution calculations.  This ensures that formulas are evaluated as they would be in `statnet` (because the evaluation happens in `statnet`).

`dissolution_coefs` just checks for `offset` and term names which can be done without getting into details of argument evaluation.  It might be a little nicer to use e.g. `list_rhs.formula` and R's `name`, `call` classes rather than strings (and maybe add more informative error messages in some cases).

As for `diss_check` I'm not aware of practically checkable necessary and sufficient conditions for (parts of) two formulas to yield the same statistics for all possible networks.  (Note that when using the "nested" EDA, we need the models to be nested not just for the initial network used in `netest` but also for all networks that could come up in `netsim`, where the network may vary from timestep to timestep in basically arbitrary ways, including changing node sets and, in particular, changing nodal attributes.)  It may be sufficient for the environments to be the same and the corresponding terms to be the same as `name`s or `call`s (after removing the `offset` decorator if present); I'll think about that a bit more, but off the top of my head I don't see a way to circumvent equivalence in that case (assuming the same `term.options` and `...` are used when initializing the models).  Those conditions are not necessary but would likely be satisfied in typical use cases.  If these checkable, sufficient conditions are not satisfied we could fall back on the non-nested EDA (perhaps giving the user an option to force the nested form if they're sure nesting is satisfied).

smjenness:
Thanks @chad-klumb for this analysis. Look forward to discussing further with you when you get to this.

chad-klumb:
I think this issue has been adequately addressed.  The netdx changes were made awhile ago.  The proposed diss_check changes would not be compatible with how EpiModelHIV-p currently produces dissolution formulas (which gives them different environments than the formation formulas in e.g. the EpiModelHIV-p tests), so I think those proposed changes would be more trouble than they're worth.  And the current dissolution_coefs error messages look reasonable to me.  So, overall, I think things are fine in their current state, and I'm closing this as completed.

Issue #652, opened 2022-04-12
consider refactoring simulation calls in netsim and/or netdx


chad-klumb:
We have 11 in `net.mod.simnet.R` and 12 in `netdx.R`.  This seems like more than should be necessary, given the number of fundamentally distinct cases we need to handle.

In `simnet`, the `tergm` and `tergmLite` cases could be handled together by simply defining the inputs and outputs appropriately.  (For single network models, as in core EpiModel, network stats can be extracted during simulation in the `tergmLite` case.)

For `netdx`, this refactoring could be part of larger changes made for #490.

Issue #490, opened 2020-12-18
Benchmark, improve netdx efficiency for parallel, big network sims


smjenness:
It seems like a lot more computational time is spent on merging and post-processing the data rather than simulating the networks here (a ratio of 5x-10x for a 25k node network):

```r
        cluster.size <- min(nsims, ncores)
        registerDoParallel(cluster.size)

        diag.sim <- foreach(i = 1:nsims) %dopar% {
          fit.sim <- simulate(fit, basis = fit$newnetwork,
                              control = set.control.ergm)
          simulate(fit.sim,
                   formation = formation,
                   dissolution = dissolution,
                   coef.form = coef.form,
                   coef.diss = coef.diss$coef.crude,
                   time.slices = nsteps,
                   constraints = constraints,
                   monitor = nwstats.formula,
                   nsim = 1,
                   control = set.control.stergm)
        }
```

Check with benchmarking more formally, and see what improvements are possible.

AdrienLeGuillou:
I benchmarked the code above with this test:

```r
pkgload::load_all()

n <- 25 * 1e3

nw <- network_initialize(n = n)
est <- netest(nw, formation = ~edges, target.stats = n / 2,
  coef.diss = dissolution_coefs(~offset(edges), 10, 0),
  edapprox = FALSE, verbose = FALSE)


profvis::profvis({
  dx <- netdx(est, nsims = 8, ncores = 8, nsteps = 100, 
              verbose = FALSE, dynamic = TRUE)
})
```

I also tried without parallelization.

When `ncores` is set to 8, the post processing, especially this call to `as.data.frame` https://github.com/EpiModel/EpiModel/blob/main/R/netdx.R#L294 takes 17 secondes out of the 24 total.

When `ncores` is set to 1, this same call still takes 17 seconds but the simulation being run sequentially take 100 seconds.

This is because `as.data.frame` is always run after the simulations are done with `lapply` so always sequentially.

A solution to speed this up would be to refactor `netdx` with a function to make the diagnostics for 1 simulation, model simulation and post processing and then either run it multiple times sequentially or in parallel. This would be similar to what is now in `netsim` with the `netsim_loop` function doing the simulation for each replication and `netsim` handling whether the replications are run sequentially or in parallel.

chad-klumb:
[...]

PR #693, opened 2022-05-23
heterogeneous dissolution updates

chad-klumb:
PR for heterogeneous dissolution updates and related changes.

Benchmarked and profiled `main` and `casl` models in the `BigNets` repository.  Each simulation is dynamic and 500 time steps.  Parallel runs use 10 cores for 30 total simulations (3 simulations per core), single runs use 1 core for 1 simulation.  Dissolution statistics are computed for `main` and `netdx_updates` branches but (being heterogeneous) not for CRAN.  `main` and `netdx_updates` branches also use `trim_netest` (CRAN does not).

|  | CRAN | `main` | `netdx_updates` |
|-|--------|-------|-----------------|
| parallel `main` | 100s | 222s | 44s |
| parallel `casl` | 174s | 419s | 70s |
| single `main` | 23s | 23s | 14s |
| single `casl` | 33s | 37s | 21s |

Profiling indicates upwards of 90% of the `netdx_updates` runtime is spent in simulation, so there doesn't appear to be much further room for improvement at the level of `EpiModel`, at least for these examples.

Dissolution model/formula evaluation (including e.g. coefficient names) is handled entirely in `statnet`.  `EpiModel` processes the change statistics returned by `tergm.godfather` to determine dissolution statistics (and, in particular, dissolution model dyad type).  Corrections for censored edges are now imputed at the level of individual edges (which, in particular, introduces variation where there previously was none).

Time indexing in the "full `tergm`" case of `netsim` was adjusted to be consistent with `netdx`.  Simulation calls in `netdx` were condensed, and default control arguments were added to `netdx`.

There don't appear to be any remaining uses of `edgelist_meanage` in core `EpiModel`, so if it's not used elsewhere it could be removed and save ~100 lines in `net.utils.R`.

references #684, #652, #490


chad-klumb:
Since `edgelist_meanage` wasn't exported, no one else should have been using it, so I've gone ahead and deleted it from this branch.  (It will remain in the github history, if we need it later.)

smjenness:
@chad-klumb : is this PR ready to merge. I can't recall if you had additional work on this to complete.

chad-klumb:
This should be ready to merge.

PR #699, opened 2022-05-25
Removes sort of redundant keep.fit arg in netest

### LINKED ISSUE #696: Default arg for netest::keep.fit
@chad-klumb : is it ok if we keep the default of netest::keep.fit to TRUE for now (perhaps changing in a later version)? If advanced users have not modified their initialization/net resimulation functions to use the formula method, then this introduces an error when the ergms are refit.

smjenness:
Also reorgs up docs in `trim_netest`

Closes #696

### DIFF
--- a/R/netest.R
+++ b/R/netest.R
@@ -30,8 +30,6 @@
 #' @param verbose If \code{TRUE}, print model fitting progress to console.
 #' @param nested.edapprox Logical. If \code{edapprox = TRUE} the dissolution
 #'        model is an initial segment of the formation model (see details).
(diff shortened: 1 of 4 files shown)


smjenness:
@chad-klumb : can you confirm [82c006d](https://github.com/EpiModel/EpiModel/pull/699/commits/82c006d094271bba5a34e0f2c78b5c13b7ef22ac) is the correct update to this unit test given the removal of `keep.fit` from netest?

chad-klumb:
should be fine; you could also leave `as.networkLite = TRUE` (the default) in the first `trim_netest` call, and delete the earlier line `est$newnetwork <- as.networkLite(est$newnetwork)`

Issue #698, opened 2022-05-24
stochastic test failure in main branch


chad-klumb:
The following test fails stochastically in `main`:

https://github.com/EpiModel/EpiModel/blob/f912380c8f2cad280e72663d10ce090fe8ef5f29/tests/testthat/test-netdx.R#L317-L363

At a glance the issue appears to be that some dissolution dyad type edge counts will occasionally hit zero, which causes plotting issues (similar to #695).

sgoodreau:
@chad-klumb I see you didn't assign anyone for this one or #695.  If you want to put me on it I'm happy to do so just tag me.

chad-klumb:
Thanks I will look into this a bit and follow up next week.

chad-klumb:
I started looking at this and #695 and I think it will be easier to fix these "well" if we first refactor the plotting code, which was already its own issue #662.  I've begun the process of refactoring in 5abdba54d4f225c40055d330bdb7ca2c0aa56088.  It will take a few more passes to get the remaining low-hanging fruit and clean up what's left; once that's done and tested we can finish addressing NA/NaN data (in what should be a relatively small number of places).

chad-klumb:
closed by #693

PR #707, opened 2022-06-20
Netsim checkpointing system

AdrienLeGuillou:
Splitting of `netsim` into subfunctions and implementation of a checkpointing system

EpiModel/EpiModelHPC#22


AdrienLeGuillou:
List of changes:

- Update the GitHub actions to v2 to prevent dependency installation failures 
- in "test-netdx.R" remove the `sim.lines = TRUE` argument on `plot(dx12)` to prevent a random error on Linux only (maybe related to #698)
-  split `netsim` internal code in several sub functions
- addition of `netsim_validate_control` that sets default values for some controls if not provided. This removes the need to do `if (!is.null(control$xxx))` each time we need them
- addition of an `on.exit` call on `netsim` to stop the `doParallel` cluster on exit
- implementation of checkpoints in netsims

smjenness (approved):

PR #740, opened 2022-07-12
netest default controls

chad-klumb:
I previously added default control arguments to `netdx`, in part to illustrate the required control classes.  This does the same for `netest`, and updates the NEWS file.


smjenness:
That's helpful, thanks!

PR #751, opened 2022-07-19
Fix the parallel memory leak

AdrienLeGuillou:
do all parallel activity in one call


smjenness:
@AdrienLeGuillou : please fix these: https://github.com/EpiModel/EpiModel/runs/7416569905?check_suite_focus=true#step:5:15

smjenness (approved):

Issue #757, opened 2022-07-27
Standardization of nwstats initialization and updating code


smjenness:
As detailed here (https://github.com/EpiModel/EpiModelHIV-p/pull/147), there has been some discrepancy in how the lists of nwstats are initialized and saved across core EpiModel and extension packages. There are different methods used if tergmLite is used, which may not be necessary. @AdrienLeGuillou additionally suggested that we could gain some efficiency if we stored all the nwstats in lists during the simulation and then bound them together at the end (i.e., in saveout.net).

 Some suggested sub-tasks:

- [ ] Review and standardize methods for initialization and updating of nwstats by core EpiModel and extension packages
- [ ] Create wrapper functions for nwstats initialization and updating that can be used across core EpiModel and extension packages. These should work for both one-layer (default) and multi-layer networks.
- [ ] Update EpiModelHIV and EpiModelCOVID with these
- [ ] In core EpiModel move the nwstats initialization from `sim_nets_t1` and into top level initialization.net function, similar to extension packages, to clarify this step.

smjenness:
@AdrienLeGuillou Could you remind us the name of the tidyverse function to rbind lists efficiently?

AdrienLeGuillou:
The function is `dplyr::bind_rows`

chad-klumb:
Initialization should be extremely simple, if it is needed at all (i.e., it may be possible to write the code in such a way that a `NULL` initial value works correctly with the general one-timestep update).

To work with either one or multiple networks, producing end-of-timestep statistics, updates to nwstats will need to be obtained through `summary` calls after all the `simulate` calls and corresponding nodal attribute updates have been performed; this could be done either at the end of the network resimulation module (calling a separately defined function so as to be reusable across EpiModel and EpiModelHIV), or as a separate module called after it (with that module itself being reusable across EpiModel and EpiModelHIV).  I probably prefer the latter option; resimulation and summary are conceptually different things.

Making these changes would be much cleaner after the `simnet_updates` and `remove_deprecated` branches are merged into `main`, which could hopefully be done shortly after our next monthly meeting.

smjenness:
Sounds great. Yes, I would like to discuss whether we need end-of-timestep stats at the next meeting.

Issue #669, opened 2022-04-26
Allow core EpiModel to handle multiple networks


AdrienLeGuillou:
#667 

This was discussed several times in the past but I can't find the original issue.

The EpiModelCOVID already have a pretty simple way of dealing with this.

One of the major issue will be to handle a step between each network resimulation, as we do with EpiModelHIV where some information are pulled from the `main` network and used in the `casl` and `oof` networks.
This can be solved with a function (or a list of them) that could be passed in `param` or `control` and would be executed in between each network simulation. (such function would be of the form `function(dat) {...; return(dat)}`)

chad-klumb:
I was asked to open a discussion on this topic, but we already have issues #667 and #669, so I'm starting by posting here.  It looks like Adrien has already been assigned to these.

@smjenness do you have more specific information as to what exactly you want to support in core EpiModel regarding multiple networks, and/or who you want to do it?

chad-klumb:
In particular:

- are we allowing node sets to differ between networks, and
- are we allowing any number of dynamic (tergm model) networks and any number of static (ergm model) networks?

smjenness:
Hi Chad - I can outline the primary issues/requests here by early next week.

smjenness:
Some summary notes following our Zoom discussion today.

1. We would like to implement a somewhat general version of multi-layer networks into core EpiModel, following the approach of EpiModelHIV as a general template. For our purposes, multi-layer networks means the same node set but different edge sets in each layer. 
2. The starting point for this would be to wrap network initialization and simulation in a loop, defined by the length of the passed in `x` object to netsim, which can be a list of separate TERGMs. Other updates with this would be to handle the multi-layer structure in the node set changes: births (arrivals) and deaths (departures). We would also move the edges correction function from the network resimulation and into these demographic functions. 
3. Implement a pre-network resimulation function that would run directly before each resimulated network. This might be a list of functions, with the same length as the length of the `x` object. This would allow a flexible way to handle what we do in EpiModelHIV: recalculate the cross-network degree values needed for updated attributes in each network simulation.
4. With step 3 working, we could consider a general multi-layer version of the infection module, following something like our [COVID model approach](https://github.com/EpiModel/EpiModelCOVID/blob/main/R/mod-infection.R#L256-L359), but one complication here is that in core EpiModel vector indexing of the infection probability parameter is already defined as time since infection. So a more general approach would be needed to potentially index by multiple factors.
[...]

PR #765, opened 2022-08-17
multinets

### LINKED ISSUE #652: consider refactoring simulation calls in netsim and/or netdx

### LINKED ISSUE #689: Remove nodefactor from diss models in future release
This functionality was deprecated starting with the merger of the netdx_heterodiss branch into main 56811b9dc20e88da5af2ea719ed2bece84592e63

### LINKED ISSUE #758: Default of nwstats.formula in multi-layer networks
This is currently an issue in extension packages (HIV, COVID) only, but probably should be handled here. With multi-layer networks, it is necessary to explicitly define the `nwstats.formula` for each layer as:

```r
nwstats.formula.1 = ...
nwstats.formula.2 = ...
nwstats.formula.3 = ...
``` 

for each layer. Nearly always we would like to use the main ERGM formula used in the model, so we input `"formation"` on the right hand side. Could this default in single layer networks be repeated for each layer in multi-layer networks so that it is not necessary to explicitly define `nwstats.formula.X = "formation"`?

### LINKED ISSUE #759: consider removing `init_tergmLite`
The (few) essential bits of it could be absorbed into `initialize.net`.  This has already been done in the `simnet_updates` branch, but we will want to discuss those changes as a whole at a later date.

Note that `EpiModelHIV` will need to be updated accordingly.

### LINKED ISSUE #760: consider removing `sim_nets_t1`
IIRC this was originally introduced to facilitate the "list of networks" approach to simulating an `ergm` in `netsim` (which has since been replaced by `tergm` simulation with `networkDynamic`s).  Much of `sim_nets_t1` is redundant with `resim_nets`, and the initial `ergm` simulation (which is not redundant with `resim_nets`) could be moved into `initialize.net`.  Some small tweaks may be needed elsewhere to make this work.

Note that `sim_nets_t1` has already been removed in the `simnet_updates` branch, but we will want to discuss those changes as a whole at a later date.

An alternative to completely removing `sim_nets_t1` may be to refactor the `tergm` simulate calls so that more code can be shared between `sim_nets_t1` and `resim_nets`.

### LINKED ISSUE #764: is it allowed to initialize a netsim run from a netsim object when `start == 1`?
In `crosscheck.net` we appear to allow this in the following check:

https://github.com/EpiModel/EpiModel/blob/8755c1e2eca75c9990a1e8cb0b4588e37eeaedf6/R/net.inputs.R#L1066-L1072

but some of the code that follows in `crosscheck.net`, as well as code in `initialize.net`, appears to assume a `netest` object when `start == 1`.

chad-klumb:
This PR combines changes refactoring `simnet`, removing deprecated functionality (`nodefactor` dissolution and `stergm` controls), and adding basic support for multiple network models in `netsim`.

Currently a draft.

closes #652 (refactoring `simnet`)

closes #689 (removing `nodefactor` dissolution)

closes #758 (adding `multilayer` control specification)

closes #759 (removing `init_tergmLite`)

closes #760 (leaving a reworked `sim_nets_t1`)

closes #764 (disallowing `start == 1` initialization from a `netsim` object)

references #669, #757


smjenness:
@chad-klumb : I have asked @AdrienLeGuillou to take a look at this work in progress, so he may have some running questions/comments we move forward.

AdrienLeGuillou:
I have some concerns regarding the network controls, I fear that the current implementation could lead to potentially hard to debug issues as `get_control` silently returns the default value if `out` is not a "multilayer".

I think it may be worthwhile to ensure that all networks have their own controls in the `dat` object. That could be inside `dat$control$network$<a list per network>` in `dat$network_controls` or something else entirely. We could then have a `get_network_control` function where the network index is mandatory. 

Such a separation would make it clear that these controls are specific to networks and to one in particular. I know that the new user that I help (and myself at the begining) have a hard time understanding what does what in the arguments we pass to control. Having a stricter structure would help to clarify things a bit. 

Also, the networks have a very unique position in EpiModel as most user rely entirely on the fact that `statnet` will work as expected (and for good reason). Therefore, having the network controls as their own thing would make sense. (Again the specifics of where we store them is not my point).

Finally, we could provide utilities like the `multilayer` class to ease the setup.

what do you think @smjenness and @chad-klumb?

NB: you know by now how I prefer simple generic structures and higher level abstraction to manipulate them over flexible functions that can handle many different situations. I believe the former is way easier to debug and update but if you think it's not the case I will not be offended :-)

chad-klumb:
Adrien, are you suggesting just a back-end change (of how we get, set, and store controls that are allowed to be specific to each network, when dealing with the `dat` object), or also a front-end change (of how those controls are specified by the user)?

I would note that making `get_control(dat, name, network = whatever)` return the full control value when `dat$control[[name]]` is not a `multilayer` was one way of addressing Sam's concern in #758 (and the corresponding issue for `set.control.*` arguments).  We can be more explicit in the back-end (at the cost of some extra code), converting the relevant controls to `multilayer` on initialization (if they are not passed in that way by the user), and then calling a more restrictive `get_network_control` accessor after that.  (We can also check that if the user does pass a `multilayer` argument then it has the correct length given the number of networks being simulated.)

AdrienLeGuillou:
my preference would be backend change and accessor changes. For the end user (i.e. the one that simply runs the final model) I don't think it matters too much. But for the one implementing modules and specific things, I would prefer to have things well separated.

I believe having a more restrictive `get_network_control` (or any name) would make the code easier to understand especially when debugging. My preference is always to have more simple functions, especially in the API part. For the final user I understand the benefice of a simple interface with flags.

And I am completely on board with the `multilayer` class that allow to check the correctness of settings to be passed to `control.net`.

chad-klumb:
updated along the lines we discussed... let me know if you have any more input!

chad-klumb:
note that network size and vertex attributes continue to be drawn from the first network when multiple networks are used

I've updated the documentation to emphasize this

AdrienLeGuillou:
thanks for the changes @chad-klumb. 

I have another proposition but I am unsure if it's the best way to go or if it is in the scope of this PR. So I am asking for your opinion (and maybe @smjenness 's) and not requesting anything:

What do you thing of having (even in mid / long term) a `dat$network` sub-list that would contain everything that is relevant to the networks? I see it as a list of size `number of networks` and for each network we would have its `control`, `el`, `nwparam`, etc.

The idea is to view each network as it's own sub object which would clearly identify what refers to it and what it impacts. 

Again, I understand that it can be a lot of code to change and that its not even desirable, so don't hesitate to disagree :-)

R/net.accessor.R
```diff
@@ -438,6 +438,35 @@ get_control <- function(dat, item, override.null.error = FALSE) {
   return(out)
 }
 
+#' @rdname net-accessor
+#' @param network index of network for which to get control
+#' @export
+get_network_control <- function(dat, item, override.null.error = FALSE, network) {
```
AdrienLeGuillou, line 444: `network` could be the second argument for consistency with the other functions with a mandatory `network` argument

R/net.accessor.R
```diff
@@ -438,6 +438,35 @@ get_control <- function(dat, item, override.null.error = FALSE) {
...
+  if (missing(network)) {
+    stop("`get_network_control` requires `network` argument.")
+  }
```
AdrienLeGuillou, line 447: `network` being a mandatory argument, we should not handle it with `missing` and instead let R crash in the `normal` way if it's not provided.

Also, it is usually agreed that `missing` should be avoid in general and replaced with a default value. Usually, `network = NULL` in the function signature and `if (is.null(network))` makes it immediately visible that the `network` argument is optional without the need to read the body of the function or the documentation.

R/net.mod.infection.R
```diff
@@ -56,7 +56,7 @@ infection.net <- function(dat, at) {
   if (nElig > 0 && nElig < nActive) {
 
     # Get discordant edgelist
-    del <- discord_edgelist(dat, at, network = 1)
+    del <- do.call(rbind, lapply(seq_along(dat$nwparam), discord_edgelist, dat = dat, at = at, include.network = TRUE))
```
AdrienLeGuillou, line 59: now that `dplyr` is imported by `EpiModel`, we can use `dplyr::bind_rows` instead of `do.call(rbind, list_of_dataframe)`. It makes for an easier to read code and it's way faster.

the code could be:

```r
del_lists <- lapply(seq_along(dat$nwparam), discord_edgelist, dat = dat, at = at, include.network = TRUE)
del <- dplyr::bind_rows(del_lists)
```

R/net.mod.infection.R
```diff
@@ -178,7 +178,7 @@ infection.2g.net <- function(dat, at) {
...
```
AdrienLeGuillou, line 181: same comment

R/net.mod.infection.R
```diff
@@ -320,6 +322,10 @@ discord_edgelist <- function(dat, at, network = 1, infstat = "i") {
     }
   }
 
+  if (NROW(del) > 0 && include.network == TRUE) {
+    del <- cbind(del, network = network)
```
AdrienLeGuillou, line 326: `cbind` can be replaced with `dplyr::bind_cols` which is faster

R/net.accessor.R
chad-klumb: I don't think R's default handling of a missing argument used as an index in a `list` is really what we want here.  For example, if we define

```
f <- function(a,b) a[[b]]
```

then

```
f(a = list(1,2,3))
```

evaluates to `NULL`; there is no error thrown, even though `b` is missing and is used to index `a`.

We can restore an error message without an explicit `missing` check by calling `force(b)` before returning `a[[b]]`.

chad-klumb: the `force` mechanism is used in dbdaab3972fab3d4a912d9ed89b844ebb50e0080 but I will leave this open for now in case there is more to discuss

AdrienLeGuillou: Oh I did not know this case! Thanks.

In this case I can see that `missing` is actually the right call, sorry for this comment

R/net.mod.infection.R
chad-klumb: addressed in 17e2faf3895ebb01e215ec77b7351b204aa2d099

chad-klumb:
In principle it might be worth considering some restructuring of the `dat` object, but I do think that is beyond the scope of this PR and should be addressed as a separate issue.

chad-klumb:
I'll be giving this some final review over the weekend and will convert it to an open PR when I'm done with that.

chad-klumb:
closes #669

closes #757

closes #767

chad-klumb:
Added `multilayer` support for `tergmLite.track.duration` as you may wish to track duration for some networks and not others. 
 (Note that tracking duration for large or dense networks with short edge durations may be relatively costly, as `lasttoggle` will accumulate many changes.)  The default remains `FALSE` for all networks.

chad-klumb:
Ready for testing/review.  Pairs with `EpiModel/EpiModelHIV-p@multinets`.

chad-klumb:
There are example usages in

https://github.com/EpiModel/EpiModel/blob/multinets/tests/testthat/test-multinets.R

and

https://github.com/EpiModel/EpiModelHIV-p/blob/multinets/tests/testthat/test-model.R

R/net.inputs.R
```diff
@@ -1033,9 +1049,13 @@ crosscheck.net <- function(x, param, init, control) {
...
 
       # Main class check ----------------------------------------------------
-      if (!inherits(x, c("netest", "netsim"))) {
-        stop("x must be either an object of class netest or class netsim",
-             call. = FALSE)
+      if (inherits(x, "netest")) {
+        x <- list(x)
+      }
+      if (!inherits(x, "list") || length(x) == 0 ||
+          !all(unlist(lapply(x, inherits, "netest")))) {
```
AdrienLeGuillou, line 1056: could we use 
`vapply(x, inherits, "netest", logical(1))` instead?

R/net.mod.init.R
```diff
@@ -23,18 +27,13 @@ initialize.net <- function(x, param, init, control, s) {
...
-    dat$nwparam <- list()
-    dat$nwparam[[1]] <- x[!(names(x) %in% c("fit", "newnetwork"))]
-
-    # Initial Network Simulation ----------------------------------------------
-
-    if (get_control(dat, "resimulate.network") == TRUE) {
-      nsteps <- 1
-    } else {
-      nsteps <- get_control(dat, "nsteps")
+    if (is(x, "netest")) {
```
AdrienLeGuillou, line 30: maybe we could use `inherits` here to future proof this?

AdrienLeGuillou:
Impressive work! Apart from the small comment above all seems good to me. I will run some tests with the EpiModelHIV branch as well and get back to you

R/net.inputs.R
chad-klumb: I've switched to `vapply` and `inherits` as suggested.

smjenness:
@chad-klumb : could you summarize which statnet packages currently in the DESCRIPTION::remotes field need to get published on CRAN for this branch to be CRAN-ready? Thanks!

chad-klumb:
requires `tergm@master` for the bug fix in statnet/tergm#99

smjenness:
Thanks @chad-klumb! We can/will review these changes at our meeting upcoming on Thursday. 

One idea in advance: would you be willing/able to write up a short working example as an addition to the EpiModel Gallery (https://github.com/EpiModel/EpiModel-Gallery)? You can see the other examples there as templates. The general idea is to provide a little more details about the implementation than is available in the function docs and/or unit tests.

chad-klumb:
I can construct a gallery example, probably after our Thursday meeting (there are examples in the tests and in `EpiModelHIV-p@multinets` that we can look at on Thursday).

chad-klumb:
This should be ready for review/merge.

AdrienLeGuillou (commented):
Well done @chad-klumb, my main comment is about the form of the `dat.updates` functions. 
I tested with EMHIV-p as well with the template and it works correctly. I will make a specific review for this code as well

R/net.inputs.R
```diff
@@ -727,17 +730,32 @@ init.net <- function(i.num, r.num, i.num.g2, r.num.g2,
...
-#'        help file for [`netdx`] for details and examples on specifying this parameter.
-#' @param set.control.stergm Deprecated control argument of class `control.simulate.network`. Use
-#'        `set.control.tergm` instead.
+#'        help file for [`netdx`] for details and examples on specifying this parameter. Supports
+#'        [`multilayer`] specification.
 #' @param save.diss.stats If `TRUE`, `netsim` will compute and save duration and dissolution
 #'        statistics for plotting and printing, provided `save.network` is `TRUE`, `tergmLite` is
 #'        `FALSE`, and the dissolution model is homogeneous.
+#' @param dat.updates Either `NULL`, a single function taking arguments `dat`,
+#'        `at`, and `network`, or a list of functions of length one greater
```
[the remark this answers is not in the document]
chad-klumb: Currently `dat.updates` uses `at = 0L` to indicate the initial cross-sectional simulation, and `at = 1L` to indicate the first time step simulation (both taking place in `sim_nets_t1`). If we relied on `get_current_timestep` with the way things are currently set up, I think they would both come back as `1L` (so the user couldn't distinguish between the initial `ergm` simulation and the first timestep `tergm` simulation).

AdrienLeGuillou:
@smjenness @chad-klumb 
just to let you know that I re-ran an EpiModelHIV-Template base project with this branch as well with the [companion EpiModelHIV-p one](https://github.com/EpiModel/EpiModel/pull/765) and it worked perfectly both for the local and HPC runs using slurmworkflows

smjenness:
Mostly for my own notes here.... There are multiple layers of merges and publications necessary here.

1. Publish `networkLite` to CRAN. `networkLite@main` contains the merged EpiModel/networkLite#8.
2. Once `networkLite` on CRAN, EpiModel/EpiModel#828 should be updated to remove the Github remote, then merged.
3. Then EpiModel/EpiModel#824 can be merged (comments/review from @AdrienLeGuillou are requested)
4. Then EpiModel/EpiModel#765 can be merged (the main multinets branch)
5. Publish EpiModel to CRAN with this. Probably will go to v2.4.0 given the size of the changes.
6.

chad-klumb:
It does look like EpiModelCOVID is out of date with some of the changes in `multinets` (and stacked PRs).  I'll open an issue to address that.

chad-klumb:
The other merges you've mentioned look comprehensive to me (given the current state of things).

smjenness (approved):

PR #766, opened 2022-08-25
update ess calculation in saveout.net to work with multiple networks

chad-klumb:
Making this as a separate PR since `multinets` itself may not be merged into `main` for some time.

references EpiModel/EpiModelHIV-p#166


chad-klumb:
I would recommend merging this into `main` to support `EpiModelHIV` applications using `saveout.net` while we continue to work on `EpiModel@multinets`.

smjenness (approved):

PR #768, opened 2022-08-31
use scenario tools for random params

### LINKED ISSUE #637: fortify the random parameters implementation with the scenario one
Scenarios and random parameter set use the same `data.frame` format. The implementation of random parameters (especially `get_param_set` should internally use the helpers `flatten_params` and `unfaltten_params`. 
This would make the implementation easier to maintain and extend

AdrienLeGuillou:
- fixes #637


smjenness:
Is this all "under the hood" work, or is there any change to the user facing interface?

smjenness:
Also, note the lintr checks: https://github.com/EpiModel/EpiModel/runs/8114619036?check_suite_focus=true#step:5:15

AdrienLeGuillou:
this is purely under the hood, I formalized the flat parameter format in the scenarios PR and this PR re-uses theses functions in `generate_random_params`.

It was one of the low hanging fruit issue that could be tackled before the next release

smjenness (approved):

PR #769, opened 2022-09-02
Add a `deduplicate` argument to `get_transmat`

AdrienLeGuillou:
optional argument to chose a single "infectious act" if multiple could
occur at the same time step.


AdrienLeGuillou:
I had to fix an error in `saveout.net` as well were the transmat was saved differently for the first sim

NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 ### NEW FEATURES
 
+- `get_transmat` now gets an optional `deduplicate` argument to chose one "infectious" acts in case multiple occured at the same time step
```
smjenness, line 5:
```suggestion
- `get_transmat` adds a `deduplicate` argument to select one transmitting acts in the case that multiple transmissions occurred at the same time step to a newly infected person.
```

smjenness:
@AdrienLeGuillou : note the R CMD Check NOTE related to the reference to the transmat columns in group_by. We may just need to artificially initialize these with `at <- NULL`. 

https://github.com/EpiModel/EpiModel/actions/runs/3053609028/jobs/4924449409#step:6:96

AdrienLeGuillou:
`at` here refers to a column in the data.frame. But I think I can quote the column names and it will work as well

AdrienLeGuillou:
I found a solution here https://dplyr.tidyverse.org/articles/programming.html#eliminating-r-cmd-check-notes using `rlang`'s `.data` object. I am testing it locally and will commit if it works

smjenness (approved):

PR #773, opened 2022-09-09
Fixes minor issues for working with observed network data

smjenness:
Fixes issues that prevented this EpiModel Gallery example from running:
https://github.com/EpiModel/EpiModel-Gallery/tree/main/2018-08-ObservedNetworkData

* `get_edgelist` stop check no longer uses `dat$param`
* `resim_nets` avoids unnecessary pulls of control settings if `resimulate.network = FALSE`

### DIFF
--- a/R/el_cuml.R
+++ b/R/el_cuml.R
@@ -15,13 +15,16 @@
 #'
 #' @export
 get_edgelist <- function(dat, network) {
-  if (!network %in% seq_along(dat[["nwparam"]])) {
-    stop("There is no network '", network, "' to get the edgelist from")
-  }
 
   if (get_control(dat, "tergmLite")) {
+    if (!network %in% seq_along(dat[["el"]])) {
+      stop("There is no network '", network, "' to get an edgelist from")
+    }
     el <- dat[["el"]][[network]]
   } else {
+    if (!network %in% seq_along(dat[["nw"]])) {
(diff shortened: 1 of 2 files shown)


chad-klumb:
I'll take a look at this over the weekend.

AdrienLeGuillou (approved):

chad-klumb:
Looks good.  Couple of thoughts for related future work:

- the number of networks is a pretty important thing so I think we should introduce a more explicit `num_nw` (or similarly named) field on the `dat` object and use that directly instead of the length of `dat$nwparam, dat$nw, dat$el` etc.

- is the intention for everything in `netsim` to keep working if we start with a positive number of nodes, drop down to zero nodes, and then increase to a positive number of nodes again?  If so, do we have any tests confirming that everything works?  Obviously this is a corner case (and not relevant when `resimulate.network = FALSE`), but perhaps it's something to look into or add a test for in the future (if there isn't one already).

chad-klumb (approved):

smjenness:
@chad-klumb : thanks for the notes above. I agree about a more solidly defined number of networks variable. I'll open an issue on this. 

And yes, the goal is for the simulation to still run even if the number of nodes drops to zero. But we need to skip the network resimulation in this case, as it is both unnecessary and triggers expected but weird errors in ergm/tergm. Definitely an edge case, but something that could arise with a catastrophic epidemic scenario.

PR #784, opened 2022-09-29
fix rng and scenarios tests

AdrienLeGuillou:
- some of the rng params test where malformed they were adding duplicated params where they shouldn't. Triggering warnings that are tested elsewhere
- the check_scenario_df function was letting a warning pass due to poor ordering of the checks


AdrienLeGuillou:
adding this PR to `unit-test-cleanup` to avoid conflict with your changes @smjenness

smjenness (approved):

Issue #750, opened 2022-07-18
Remove nolint tags when next version of lintr is published


smjenness:
@AdrienLeGuillou : I bet you know the answer to this one. `lintr` picks up a "no visible binding" issue, like below:

<img width="643" alt="Screen Shot 2022-07-17 at 8 01 42 PM" src="https://user-images.githubusercontent.com/2521570/179430202-cdc7098f-4f26-49b9-9ecb-85640fd01918.png">

in all cases in EpiModel with we use the `with` function to extract columns from data frames (it is also picked up in your attributes and summary stats vignette for the `status` variable that also does this).

Are we using improper syntax here, or is this just a ignorable lintr issue?

smjenness:
Ignored for now with `# nolint` tags in the scripts. Keeping this issue open in case the lintr team fixes this issue at some point.

AdrienLeGuillou:
It is now resolved on `lintr` by https://github.com/r-lib/lintr/pull/1548 and the next release (>3.0.1) should contain it.

AdrienLeGuillou:
yes, I suscribed to the release notification of `lintr` not to miss it

smjenness:
@AdrienLeGuillou : can this be closed?

AdrienLeGuillou:
I tried to use the new `skip_with` argument but it is not available. #863

PR #788, opened 2022-10-21
remove nolint in summary.R for

AdrienLeGuillou:
#750


AdrienLeGuillou:
indeed,
I just added the flag to ".lintr", theoretically this should be the default behavior now but at list on my machine it's now fixed

AdrienLeGuillou:
its seems there is an issue on `lintr` the 3.0.2 commits suggest that the option was added but in practice the source code is not up to date. 

Closing this and waiting on response from upstream

Issue #790, opened 2022-11-03
prevent crashes on `saveout.net` if some elements do no exist


AdrienLeGuillou:
related to #789

AdrienLeGuillou:
fixed by #791

chad-klumb:
Just a thought, but since `NULL` could be a valid value we actually want to save, should we consider replacing things like `out[[el.name]][[s]] <- dat[[el.name]]` with `out[[el.name]] <- c(out[[el.name]], list(dat[[el.name]]))`?  The latter appends `dat[[el.name]]` to `out[[el.name]]` even if `dat[[el.name]]` is `NULL`.

AdrienLeGuillou:
I see your point, I opened a new issue so I won't forget about it

PR #791, opened 2022-11-04
Prevent errors when missing elements are saved

AdrienLeGuillou:
Changes the behavior of `saveout.net` on the last sim of a batch. It now tries to rename the elements if there number match the number of simulation. In case of missmatch, a warning is produced.

addresses #790


smjenness:
I think this change merits a unit test, plus an update to NEWS.

smjenness (changes requested):
I think this change merits a unit test, plus an update to NEWS.

AdrienLeGuillou:
I added a unit test for the new function.
The integration tests were already there as the output is not changing

PR #799, opened 2022-12-06
low level logic to disable modules

AdrienLeGuillou:
WIP

- makes a named list of module on initialization
- add a "utility module" afeter `updater_net` that will disable modules listed in `param$.modules.to.disable`
- this param is cleared after the modules are disabled
- meant to be used with scenarios to chose at what time to remove what modules

### DIFF
--- a/R/netsim.R
+++ b/R/netsim.R
@@ -205,6 +205,7 @@ netsim_initialize <- function(x, param, init, control, s = 1) {
   } else {
     param <- generate_random_params(param, verbose = FALSE)
     dat <- control[["initialize.FUN"]](x, param, init, control, s)
+    dat <- make_module_list(dat)
     if (get_control(dat, "start") != 1) {
       dat <- set_current_timestep(dat, get_control(dat, "start") - 1)
     }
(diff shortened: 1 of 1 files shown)


AdrienLeGuillou:
Current state:
- the core elements are working
- user facing interface to finalize

after discussion with @smjenness, revised specs:
- do not define the `.modules.to.remove` through scenarios as initially planned
- pass the modules to remove to `control.net` and not as params
- (no changes to the internal code though)

AdrienLeGuillou:
This PR creates a list of modules in the initialization step to allow removing some of them later on in the simulation. 

The new `end.horizon` control will trigger the removal of a module set at a given time step

To trigger an end horizon at step 204 that would remove the arrival and infection modules, one would do:

```r
control.net(
  ...,
  end.horizon = list(at = 204, modules = c("arrivals.FUN", "infections.FUN"),
  ...
)
```

At initiation, a check verifies that the modules to be removed are actually in the module list and that the time step of trigger is after the start of simulation.

@smjenness: was that the type of user facing interface you wanted? I can easily change it if needed

smjenness (changes requested):
This is really great, Adrien! My only minor comment is that the unit testing could be strengthened. For example, if we drop the network resimulation module at time step 5, then we should expect the number of edges to stay constant from 5 to nsteps. That is something you can test with the netsim object with `get_nwstats`. Perhaps think about that too for the recovery module. This will just further solidify that the modules are dropped as expected. Thanks!

smjenness:
@AdrienLeGuillou : as per our conversation earlier today, I am merging in this PR now so that colleagues can start working with the functionality on the main branch, but please do open a new PR to add extra unit tests (and potentially move that nwstats calculation code from the network resimulation module to the prevalence module). Thanks!

PR #810, opened 2023-02-09
Flex modules

AdrienLeGuillou:
more tests for the module removal system:

- test the removal of 2 modules at once
- test the correct output with theses modules removed

I have not moved the `nwstats` calculation to `prevalence.FUN` yet as it may conflict with #765 (@smjenness ,should I do it anyway?)


smjenness:
@AdrienLeGuillou : I am going to merge this in soon, but the multi-layer nets is not ready for CRAN yet, so I will ask you to file an issue to move the nwstats calculation over to prevalence.NET only after we merge that in (probably in another couple weeks).

AdrienLeGuillou:
already done in #811 :)

PR #824, opened 2023-02-25
update reinitialization to handle tergmLite and EpiModelHIV-p

### LINKED ISSUE #802: support restarting tergmLite simulations in core EpiModel
The current code does not appear to handle this.  For example, it doesn't pull `el` from the original simulation.

https://github.com/EpiModel/EpiModel/blob/7aa3a975f9363a653f54fba55cac75316ba3a401/R/net.mod.init.R#L72-L90

chad-klumb:
stacked on #765

closes #802

There were some discrepancies / issues with the reinitialization functions that are worth mentioning.

- `temp` was not required or copied in core EpiModel's previous reinitialization function.  This produced errors in general when reinitializing core EpiModel simulations that used `epi.by` or open populations with attributes (since important data is stored in `temp` in those cases).  Since `temp` was always required by EpiModelHIV's reinitialization function, I have made it required in the new combined reinitialization function.  Note that core EpiModel users wishing to reinitialize their simulations now need to include `"temp"` in the `save.other` control argument, regardless of whether or not their modules actually require `temp`.  This is a bit annoying.  We do have the option of always saving `temp` in `saveout.net`, which would circumvent the issue of the user needing to include `"temp"` in `save.other`.  It looks like the `temp` values in core EpiModel tend to be small (hence not a memory issue for including in `saveout.net`), but there could conceivably be some large `temp`s in custom or extension modules.

- Core EpiModel pulled the `param` value from `x` when re-initializing, whereas EpiModelHIV used the `param` argument to `reinit_msm`.  The combined reinitialization function pulls `param` from `x`.  If there is a need to allow a different `param` argument in `EpiModelHIV` then we will want to adjust this.

- Since `tergmLite` simulation with `tergmLite.track.duration = TRUE` stores the `lasttoggle` data structure on the network in `dat`, we need to allow saving the network when `tergmLite = TRUE` in order to reinitialize correctly.  I've updated the code to allow `save.network = TRUE` when running `tergmLite` simulation; the network being saved is the `networkLite` after the final timestep.  More generally, the user might be interested in the final network state even if the simulation is a `tergmLite` simulation, so it seems reasonable to let them save the `networkLite` using `save.network = TRUE`.


smjenness:
@AdrienLeGuillou : I will ask you to review this PR, with specific focus on the first bullet point related to the use of `temp`.

PR #825, opened 2023-03-01
updates for compatibility with ergm@master

chad-klumb:
There are changes (specifically statnet/ergm@20c6c4c18f8803838243a8d08b1c1fa66a50d189) in `ergm@master` that break `EpiModel@main` by changing the availability of `newnetwork` on the `ergm` fit object.  (It's no longer available when only running the initial fit, which is often what happens by default for dyad-independent models.)  This PR updates `EpiModel@main` to remain compatible with `ergm@master` by falling back on `fit$network` when `fit$newnetwork` is `NULL`.  If/when the changes in statnet/ergm@20c6c4c18f8803838243a8d08b1c1fa66a50d189 are pushed to CRAN you will need this `EpiModel` update to avoid breakages.  I'm including `ergm@master` in the remotes specifically to confirm compatibility with these `ergm` changes.

Also taking the opportunity to remove some `networkLite` tests that can go elsewhere (a couple of which were broken by the `ergm@master` changes).


chad-klumb:
`tergm@master` is also required due to changes in `simulate` handling of the `dynamic` argument (statnet/ergm@f7c168e9f04362196353dace5ff2649bd2ad85eb, statnet/tergm@95e867e9ce6c1038b0be172097ccf35529bdbd39)

smjenness:
Thanks @chad-klumb. Should this also require an update to the minimum versions of ergm and tergm required in DESCRIPTION, or can/should that be handled once those packages are published to CRAN?

chad-klumb:
The `basis` changes do require `master` versions so I've updated the version numbers in the DESCRIPTION file.  I would update these to CRAN version numbers when the statnet `master` dependencies are pushed to CRAN.

krivit:
Would it be simpler if I just put `$newnetwork` back in unconditionally?

chad-klumb:
up to Sam really; the `basis` changes that are now part of this PR require `@master` versions of `ergm`, `tergm`, and `ergm.ego`, regardless of `newnetwork` availability

krivit:
@smjenness, thoughts?

smjenness:
@krivit : as @chad-klumb notes, he already developed an update for EpiModel that can work without `newnetwork` as there are also changes to `basis`. I'm fine with that, we just need to coordinate a release together. Let me know when you are ready for that. (see also the CRAN email about ergm.ego that I forwarded that has implications for the above too)

smjenness:
Merging with irrelevant windows R CMD Check error (pak issue)

PR #828, opened 2023-03-06
add separate net_attr dat field for tergmLite simulation

### LINKED ISSUE #801: consider giving network attributes their own `dat` field in `tergmLite` simulation
Currently network attributes in `tergmLite` simulation are stored as `attr`-style attributes on the edgelist. This is somewhat awkward, and I think it would make more sense to give them their own `dat` field, named something like `network_attributes` (or `net_attr` if `network_attributes` is too long). Like `el`, there will need to be one entry in `network_attributes` for each network, since (for example) the `lasttoggle` network attribute varies from network to network.

chad-klumb:
stacked on #824

closes #801

removes use of `dat$nw` for tergmLite simulation, storing the deconstructed network in the `el`, `attr`, and `net_attr` fields

updates other functions as needed to work with this change; in particular

- `copy_nwattr_to_datattr` now requires the network to be passed (since `dat$nw` is not available in the `tergmLite` case), and
- `init_status.net` no longer pulls `status` from `dat$nw[[1]]` when `statOnNw = TRUE`, since the status has already been copied  by an earlier call to `copy_nwattr_to_datattr` in `initialize.net`, and
- the saved network is no longer required for restarting tergmLite simulation (cf. #824), instead `net_attr` is (newly) required

depends on EpiModel/networkLite@net_attr branch


smjenness:
networkLite v1.0.5 now on CRAN, so this Github remote may be dropped.

chad-klumb:
done

PR #833, opened 2023-03-14
add the get_cumulative_degree function

### LINKED ISSUE #808: make a `get_degree_cuml` function
use a flag and / or duration to switch between current and "historical" degree

warning / error if duration > how much steps are stored in el_cuml

AdrienLeGuillou:
fixes #808


R/el_cuml.R
```diff
@@ -222,3 +222,30 @@ get_partners <- function(dat, index_posit_ids, networks = NULL,
...
+
+#' @title Return the Cumulative Degree of a Set of Index Nodes
+#'
+#' @inheritParams get_partners
+#'
+#' @return
+#' A \code{data.frame} with 2 columns:
+#' \itemize{
+#'   \item \code{index}: the unique ID (see \code{get_unique_ids}) of the
+#'         indexes.
```
smjenness, line 234: So this takes as an argument the posit ID but returns the unique ID. Is that necessary/justified? Is there any case where we are getting the cumulative degree for nodes without a posit ID (i.e., nodes who are already departed from the network)?

AdrienLeGuillou: Good catch!
You are right in this case it does not really make sense as we only work with information on active nodes anyways.
I will fix

smjenness: Great, thanks!

smjenness (approved):

PR #840, opened 2023-03-23
no message on step 1 for get_cumulative edgelist

### LINKED ISSUE #838: get_cumulative_edgelist message unnecessary timestep 1
@AdrienLeGuillou : I believe this is a new addition related to the requested `get_partners` warning. I find that the message generated when creating the cumulative edgelist in the init module at time step 1 unnecessary. Specifically, this is what we get when running any new model:

```r
At timestep = 1:
    the cumulative edgelist for network '1' is created from scratch 
    ** if this message repeats, check that the  
'    ** `update_cumulative_edgelist` function 
    ** is called after the networks resimulation. 



At timestep = 1:
    the cumulative edgelist for network '2' is created from scratch 
    ** if this message repeats, check that the  
'    ** `update_cumulative_edgelist` function 
    ** is called after the networks resimulation. 



At timestep = 1:
    the cumulative edgelist for network '3' is created from scratch 
    ** if this message repeats, check that the  
'    ** `update_cumulative_edgelist` function 
    ** is called after the networks resimulation. 
```

While not a bug, I find these this message is flagging a potential problem that doesn't exist in my case (i.e., I don't see the message repeating because the update function is called correctly). The warning should only appear in the case that the cumulative edgelist does not exist or not storing data correctly, so can we skip this initialization message?

AdrienLeGuillou:
fixes #838
also change for `is_null` instead of testing length. This way if only network 3 is updated, we still get the message when trying to access networks 1 and 2 as they are not there.

### DIFF
--- a/R/el_cuml.R
+++ b/R/el_cuml.R
@@ -63,7 +63,7 @@ get_cumulative_edgelist <- function(dat, network) {
       calculated. (`cumulative.edgelist` control == FALSE)")
   }
 
-  if (length(dat[["el.cuml"]]) < network) {
+  if (is.null(dat[["el.cuml"]][[network]]) && get_current_timestep(dat) > 1) {
   message(
       "\n\nAt timestep = ", get_current_timestep(dat), ":\n",
       "    the cumulative edgelist for network '", network,


AdrienLeGuillou:
closing, superseeded by #842

PR #842, opened 2023-03-23
Cleaner el cuml

AdrienLeGuillou:
put the `cumulative.edgelist` control check inside `update_cumulative_edgelist` to govern weather or not to update.

error if calling `get_cumulative_edgelist` and `cumulative.edgelist == FALSE`

This will catch all cases of trying to use cumulative edgelist when its not calculated


AdrienLeGuillou:
also, `update_cumulative_edgelist` is now called unconditionally and returns early if `cumulative.edgelist == FALSE`

AdrienLeGuillou:
note: 
`cumulative.edgelist` and `truncate.cumulative.edgelist` get values of FALSE and 0 if not set in `control.net` (through `validate_control`)

smjenness (approved):

chad-klumb:
Are there corresponding updates for EpiModelHIV-p?  (The current EpiModelHIV-p@dev branch does not seem to work with these updates to EpiModel@main.)

PR #852, opened 2023-04-05
On reinit, use provided params

AdrienLeGuillou:
The reinit code in EpiModel now superseeds the `reinit_msm` however, this function behavior is to copy the params from the restart point and not use the ones provided by netsim.

This PR reverts that as it is the  way we modify params for the scenario API (and even without).

### DIFF
--- a/R/net.mod.init.R
+++ b/R/net.mod.init.R
@@ -72,7 +72,7 @@ initialize.net <- function(x, param, init, control, s) {
            paste.and(missing_names), call. = FALSE)
     }
 
-    dat <- create_dat_object(param = x$param, control = control)
+    dat <- create_dat_object(param = param, control = control)
 
     dat$num.nw <- x$num.nw
     if (control[["tergmLite"]] == TRUE) {
--- a/R/net.mod.init.R
+++ b/R/net.mod.init.R
@@ -74,9 +74,14 @@ initialize.net <- function(x, param, init, control, s) {
 
     dat <- create_dat_object(param = param, control = control)
 
-    if (is.null(param[["groups"]])) {
-      dat <- set_param(dat, "groups", x$param$groups)
+    missing_params <- setdiff(names(x$param), names(param))
+    for (mp in missing_params) {
+      dat <- set_param(dat, mp, x$param[[mp]])
     }
+
+    # recycle sims in the restart object
+    # e.g. 5 sim out of a size 3 restart object we will give: 1, 2, 3, 1, 2
+    s <- (s - 1) %% length(x$attr) + 1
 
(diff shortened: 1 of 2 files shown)


chad-klumb:
cf. https://github.com/EpiModel/EpiModel/pull/824#issue-1599791779 for a discussion of re-initialization issues (including use of `param`)

If using the provided params you might want to look at `time.unit` in EpiModelHIV-p as well:

https://github.com/EpiModel/EpiModelHIV-p/blob/e2ed9e5c451309ef008ac0757099622b6cb851af/R/mod.initialize.R#L54-L56

https://github.com/EpiModel/EpiModelHIV-p/blob/e2ed9e5c451309ef008ac0757099622b6cb851af/R/mod.initialize.R#L428-L430

To avoid the need to manually pull it from `param$epistats` to the top level of `param` within the (re)initialization code perhaps that could be done within the `param_msm()` function itself (i.e. output `time.unit` from `epistats` as a top-level entry in the `param_msm` list)?

AdrienLeGuillou:
Thanks @chad-klumb for the link to the discussion. I definitively missed the implications of that when reviewing that PR at the time.

Using the provided params is (unless I am missing something) mandatory for the restart system to be useful. Otherwise it is just an option to pause and resume a simulation without making any changes to it. And the way we use this system in EpiModelHIV is:

- calibrate over ~3000 steps
- pick a restart point
- run the different interventions from the restart point to the end by giving new params

Maybe a simple approach could be to update the stored param list from `x` using the one provided to `netsim`.

About `param_msm`, we've been discussing with @smjenness ways we could stop using it entirely in the future to rely on `param.net` instead. So adding things there would not be my preferred approach.

AdrienLeGuillou:
As this is a longer discussion than expected, I made that into a draft and made this quickfix to EpiModelHIV-p that only allows restart with scenarios while I am away (until april 18) : https://github.com/EpiModel/EpiModelHIV-p/tree/DoxyPEP-reinit

AdrienLeGuillou:
Back to it, the las commit provides:

- a fix to restart `netsim` where the number of sim is greater than the size of the restart point. It recycles the restart object similarly to vector recycling in the rest or R
- I updated the restart test to restart 2 sim out of a one sim restart point
- reinit uses the provided params and fill the blanks with the ones from the restart. This fixes the issue with "groups" and "time.unit".

AdrienLeGuillou:
I also works with EpiModelHIV-p@DoxyPEP

@smjenness @chad-klumb do you see something I missed?

chad-klumb (approved):
Looks good to me.

smjenness (approved):

PR #868, opened 2023-07-27
Create generic get_discordant_edgelist()

### LINKED ISSUE #795: make a generic `get_discordant_edgelist` function
This function would be a more general version of `discord_edgelist`. It would simplify models with multiple diseases as well as standardize the format of the discordant edgelist. (making it related to https://github.com/EpiModel/EpiModelHIV-p/issues/129)

I envision a function signature like:

```r
get_discordant_edgelist <- function(dat, status.attr, head.status = "i",  tail.status = "s", networks = NULL)
```

where:
- `status.attr`: the name of the attribute to use `status`, `gc.status`, etc
- `head.status`: what values on `status.attr` to look for in the "head" of the edge (default "i" for SI* models)
- `tail.status`: what values on `status.attr` to look for in the "tail" of the edge (default "s" for SI* models)
- `networks`: on what network should we look

This function can use `get_partners` by giving the `posit_ids` of the "infectious" nodes then filtering out the non "susceptible". (Conversion to `posit_ids` would be required at the end)

Similar to  `get_partners`, the output could be a `tibble`:

|head|tail|network|head_status|tail_status|
|-|-|-|-|-|

### examples

```r
# HIV on EpiModelHIV
del <- get_discordant_edgelist(dat, "status", 1, 0)

# MonkeyPox (SEIR) where only 2 networks out of 3 are of interest
del <- get_discordant_edgelist(dat, "status.mpx", "i", "s", networks = c(1, 2))

# arbitrary disease where 2 status are infectious
del <- get_discordant_edgelist(dat, "arbitrary", c("i.a", "i.b"), "s")
```

karina-d-wallrafen:
Closes #795


AdrienLeGuillou (changes requested):
Hi @karina-d-wallrafen,

first of, from what I see I am pretty sure your code does the trick. Please view my comments and suggestions as a way to keep the code in a coherent state with what I build around this subject before.

The most import point I think when building these internal tools is to not hesitate to split the different parts of the function into multiple sub-function. It makes the code easier to read and therefore to debug afterwards. (and dev is rougly 85% debugging)

Below I made an example implementation of the feature so you can see what I mean.
My goal was to make simple sub-function that gives garantes about there output so we can use it without checking its validity at each step. (That's the principle of unit testing actually) 

I also made comments on specific parts of your code.

Do not hesitate if you have any questions

```r
# generic function to turn an edge list into a tibble with a fixed format
# (whatever the size of the el, the output always have the same columns even if 0 rows)
as_tibble_edgelist <- function(el) {
  if (nrow(el) > 0) {
    d_el <- tibble::tibble(head = el[, 1], tail = el[, 2])
  } else {
    d_el <- tibble::tibble(head = integer(0), tail = integer(0))
  }
  return(d_el)
}

# function to return the edgelists as a formatted tibble with a "network" column
# similar to get_cumulative_edgelists_df
# builds of as_tibble_edgelist to enforce the format
# similarily, the output tibble is garanted even if all el are empty
get_edgelists_df <- function(dat, networks = NULL) {
  networks <- if (is.null(networks)) seq_len(dat$num.nw) else networks
  el_list <- lapply(networks, get_edgelist, dat = dat)
  d_el <- dplyr::bind_rows(lapply(el_list, as_tibble_edgelist))

  el_sizes <- vapply(el_list, nrow, numeric(1))
  d_el[["network"]] <- rep(networks, el_sizes)

  return(d_el)
}

#
get_discordant_edgelists_df <- function(dat, status.attr,
                                        head.status, tail.status,
                                        networks = NULL) {
  d_el <- get_edgelists_df(dat, networks)
  status <- get_attr(dat, status.attr)

  # discordant edgelist where head and tail are correctly ordered
  d_el_direct <- dplyr::filter(
    d_el,
    status[d_el$head] %in% head.status,
    status[d_el$tail] %in% tail.status,
  )

  # discordant edgelist where head and tail are inverted (fixed in last line)
  d_el_inverse <- dplyr::filter(
    d_el,
    status[d_el$tail] %in% head.status,
    status[d_el$head] %in% tail.status,
  ) |>
  dplyr::select(head = "tail", tail = "head", "network")

  # bind, add statuses and reorder columns
  d_el <- dplyr::bind_rows(d_el_direct, d_el_inverse) |>
  dplyr::mutate(
    d_el,
    head_status = status[.data$head],
    tail_status = status[.data$tail]
  ) |>
  dplyr::select(
    d_el,
    "head", "tail",
    "head_status", "tail_status",
    "network"
  )

  return(d_el)
}
```

karina-d-wallrafen:
Thanks for the feedback @AdrienLeGuillou. I've made the changes you requested.

smjenness:
Just a note before we wrap this up that there are several lintr flags: https://github.com/EpiModel/EpiModel/actions/runs/6028675911/job/16356460908?pr=868#step:5:17

karina-d-wallrafen:
@smjenness lintr flags have been addressed!

R/net.mod.infection.R
```diff
@@ -331,6 +331,121 @@ discord_edgelist <- function(dat, at, network = 1, infstat = "i", include.networ
...
+#' @seealso \code{\link{discord_edgelist}}
+#'
+#' @export
+#' @keywords netMod internal
+#'
+get_discordant_edgelist <- function(dat, status.attr, head.status,
+                                    tail.status, networks = NULL) {
+
+  if (length(intersect(head.status, tail.status)) > 0 &&
+        get_current_timestep(dat) == (get_control(dat, "start") + 1)) {
```
AdrienLeGuillou, line 372: I think the  time condition should be first here.
The `&&` operator is short circuited which means that it will only perform the `intersect` call if at time `start + 1`. And in general, set operation (`intersect`, `setdiff`, etc) are more computationally intensive compared to getters

AdrienLeGuillou (changes requested):
Looks good and clear to me.
I think we should have these function in another file, like "R/net.edgelist.R". Maybe we could also put the code from "R/el_cuml.R" there as well as it also interact with edgelists.

Thanks for your work here!

AdrienLeGuillou (approved):

PR #872, opened 2023-09-15
Use `collection::queue` for `attr.history`

AdrienLeGuillou:
Use `queue` (linked list) instead of R `list` to store the `attr.history` elements. This data structure massivelly reduces the cost of appending new elements to `attr.history`.
The implementation is very similar to before as the `as_list()` method allow simple conversion to a list once we want to retrieve the data.

The idea to use this structure comes from this EpiModelHIV-p function: https://github.com/EpiModel/EpiModelHIV-p/blob/DoxyPEP/R/windowed_counter.R


smjenness (approved):

smjenness:
Nice idea, thank you!

PR #886, opened 2024-05-01
Add helpers to work with data.frame of parameters

AdrienLeGuillou:
Convert from and to data.frame of parameters with `param.net_from_table` and `param.net_to_table`.

Split these code to their own file and add tests.

`param.net_from_table` is part of EpiModel since the use of the `param.df` argument to `param.net` but was not exported. `param.net_to_table` is simply the reverse operation. It's useful to store calibration results back into a csv or xlsx file.


smjenness:
Great idea, thanks!

PR #892, opened 2024-07-09
Reachable nodes

AdrienLeGuillou:
Export 3 new functions:
- `get_forward_reachable` 
- `get_backward_reachable`
- `as_cumulative_edgelist.networkDynamic`

They compute the set forward and backward reachable from a subset of nodes in a dynamic network represented in a cumulative edgelist.

These functions are time and memory efficient versions of `tsna::tPath` for multiple nodes and when the only information of interest is the presence or absent of a node in the reachable set over a time period.

`get_backward_reachable` is only a wrapper around `get_forward_reachable` where the edges times have been inverted (`head = -tail` and `tail = -head`)

`as_cumulative_edgelist.networkDynamic` returns a cumulative edgelist representation of a `networkDynamic` object. This representation is lightweight and formatted similar to the other cumulative edgelists uses all over `EpiModel`.


martinamorris:
nice to see someone using these utilities :)

AdrienLeGuillou:
@martinamorris & @smjenness : do you think these new utilities should be added to EpiModel or to tsna?

arguments for EpiModel:
- they rely on the cumulative edgelist representation used in EpiModel
- probably will be used on outputs from `netsim`

arguments for tsna:
- based on the initial idea of `tPath` (and friends)
- fit the "Tools for Temporal Social Network Analysis" description
- could make sense to have a formalized `cumulative edgelist` object type there
- reduce the amount of exported stuff from EpiModel

I don't feel strongly either way, but I had a note for proposing the earlier version of these utilities to `tsna` for a long time

martinamorris:
@AdrienLeGuillou good question.

In principle, I like the idea as it seems cleaner.   I don't see tsna listed in the depends/imports for EpiModel, so that's a question for @smjenness.

smjenness:
Let's put this in EpiModel for the time being because of the potential duplication in function and default handling of time step s that differs between nD and EpiModel.

Issue #877, opened 2024-02-05
`plot.netsim()` do not start the lines where it should after a restart


AdrienLeGuillou:
after a simulation restart, when an EpiTracker is added (not present in the original simulation), `plot.netsim()` make the line start at 0 instead of the restart point.

The axis however are of the right dimensions (all timesteps)

I suppose the issue comes from a call to `na.rm` somewhere before the plot is made. As new trackers have `NA` for all the values before creation.

Note: the issue is not present with `as.data.frame` and `ggplot2`

AdrienLeGuillou:
error in `draw_means` in the `supsmu` call.
Probably a similar error with `draw_qnts` ?

Issue #670, opened 2022-04-26
replace the use of `missing` with `args = NULL` defaults


AdrienLeGuillou:
In our function calls, arguments that can be omitted should be set to `NULL` in the function signature and not rely on `missing` to check if they were provided or not. This makes it obvious that they are optional.

This change should happened at once all over the code base to avoid discrepancies. This is especially true with `param.net`, `control.net` and `init.net` that rely heavily on this approach.

AdrienLeGuillou:
Warning, this can lead to issues when the argument is used to subset elements in the function. See https://github.com/EpiModel/EpiModel/pull/765/files#r959770128

PR #903, opened 2024-08-23
plot fixes: `axs`, `...`, `NA`s in smooth

### LINKED ISSUE #859: Remove plot.netsim (and probably elsewhere) axs argument
This does not get processed correctly. `xaxs` and `yaxs` can still get passed through `...`, so it's unnecessary in any case.

AdrienLeGuillou:
remove the `axs` arguments to the plot function
and instead pass the ellipsis to the inner call to plot

fixes #859


AdrienLeGuillou:
current blockers:

the plot functions will check for the presence of some elements in the ellipsis and pass them explicitly if afterwards.
Therefore, passing the ellipsis to the inner `plot` results in a double definition of the arguments.

Solution:
Every argument checked by the function with `is.null(da$xxx)` should instead be passed explicitely with a default `NULL` value. 

This is actually similar to the use of the `missing` function. See #670

AdrienLeGuillou:
split of the `plot.R` script to avoid a 2k lines file

remove all `da$args` within `plot.icm` and `plot.dcm` replacing them with explicit params. Fixing the double argument passing above

- all `missing(xxx)` in `plot.icm` and `plot.dcm` were replaced with explicit default or `NULL` in the function calls.
- removal of some dead branches in the code

AdrienLeGuillou:
fix #877 

`supsmu` returns `x` and `y` value but only the `y` was used in `draw_means`.
The `x`s are useful when `NA`s are present as they indicate where real values are present.

AdrienLeGuillou:
superseded by #905

PR #913, opened 2024-09-18
Test whether ergm devel fixes R CMD Check error

smjenness:


AdrienLeGuillou:
it works with ergm & tergm from master:
#914

PR #929, opened 2025-04-09
allow epi data frame from truncated sims

AdrienLeGuillou:
Previously `as.epi.data.frame` required untrimmed simulations. 

This update allows making `epi.data.frame` (used for plotting) out of trimmed `data.frame`. This would typically come from `merged_tibbles`.


AdrienLeGuillou:
It actually creates more problems with the plotting function than it solves

PR #1004, opened 2026-04-09
Attributes tracking in EpiModel

AdrienLeGuillou:
### Summary

- Add automatic delta-compressed attribute history tracking to `netsim`, enabled via `tracked.attributes` in `control.net()`
- Add `get_attr_at()` to reconstruct tracked attribute state at any past time step
- Add `make_networkDynamic()` to rebuild a full `networkDynamic` object from `tergmLite` simulation output, enabling `ndtv` visualization without sacrificing simulation performance
- Add `record_raw_object()` for ad-hoc object recording during simulation (debugging/analysis)
- Fix bugs in `saveout.net`, `record_attr_history`
- Expand the "Attributes and Summary Statistics" vignette with documentation for all new features

### Motivation

I wanted a way to explore EpiModelHIV behavior in depth without storing the `attr` list at every time step. This PR offers a leaner approach to attribute tracking.

As a bonus, the tracked history can be combined with the cumulative edgelist to rebuild a `networkDynamic` object post-hoc, enabling `ndtv` animation even from fast `tergmLite` runs. (see below)

### New public API

| Function | Purpose |
|----------|---------|
| `get_attr_at(sim, at, sim_num)` | Reconstruct a tibble of tracked attributes for all active nodes at time step `at` |
| `make_networkDynamic(sim, sim_num, network)` | Rebuild a `networkDynamic` from cumulative edgelist + tracked attribute history |

The existing `record_attr_history()` function can still be called manually inside custom modules for ad-hoc recording alongside automatic tracking.

### How it works

**Setup.** The user lists attribute names in `control.net(tracked.attributes = c("status", ...))`. The `active` attribute is always included automatically.

**At each time step**, the simulation pipeline:
1. Snapshots tracked attributes before modules run (`tracked_attrs_set_ref`)
2. Runs the module pipeline (infection, recovery, departures, arrivals, prevalence, ...)
3. Diffs current state against the snapshot and records only changes (`tracked_attrs_record`)

Three types of changes are detected:
- **Departures**: nodes present in the snapshot but absent now (recorded as `active = 0`)
- **Arrivals**: nodes absent from the snapshot but present now (all tracked values recorded)
- **Value changes**: persisting nodes whose attribute value differs

On the first call, a full initial snapshot is recorded instead of a delta.

**Post-simulation**, `get_attr_at(sim, at = 50)` replays the history forward to reconstruct the complete attribute state at any time step. `make_networkDynamic()` combines this with the cumulative edgelist to produce a full `networkDynamic` suitable for `ndtv::render.d3movie()`.

### Design decisions reminders

- **UIDs are always contiguous starting from 1.** `make_networkDynamic` uses `max(uids)` as the network size and UIDs directly as vertex indices. This is guaranteed by EpiModel's UID allocation.
- **Nodes cannot re-enter.** `get_nodes_spell()` computes a single onset/terminus pair per node using `min()`. Once a node departs (`active = 0`), it stays departed. This is an invariant of EpiModel's demographic model.
- **`get_attr_at` always includes `active`.** The `active` column is included in the output tibble for consistency, even though all returned nodes are active by definition (they passed the spell filter). This matches the convention of attribute accessors elsewhere in EpiModel.
- **Departures usually happen early in a time step** (before infection/recovery). So `at < terminus` in `get_attr_at` consider only nodes present at the end of a step. Consistent with epi trackers.
- **Checkpoint keeps history; restart does not.** When using `.checkpoint.steps`, the full `dat` object (including `attr.history`) is serialized, so checkpointed simulations preserve history across interruptions. When restarting from a prior `netsim` output (`control$start > 1`), history is intentionally not carried over: the restart creates a fresh `dat` and records a new initial snapshot at the restart time.
- **`attr.history` is always saved in output**, even when empty (no tracking configured). This is consistent with the unconditional save of `raw.records` and simplifies downstream code that accesses these fields.

### Example: animating a `tergmLite` simulation with `ndtv`

This was previously impossible without running the full (slow) `networkDynamic` mode.

```r
library(EpiModel)
library(ndtv)

# works with bigger networks and longer simulations, but the video becomes ugly
n <- 1e2
nw <- network_initialize(n = n)
formation <- ~edges
target.stats <- 0.4 * n
coef.diss <- dissolution_coefs(dissolution = ~ offset(edges), duration = 10)
est <- netest(nw, formation, target.stats, coef.diss)

# 2. Epidemic simulation
param <- param.net(inf.prob = 1)
init <- init.net(i.num = 0.1 * n)
control <- control.net(
  type = "SI",
  nsteps = 100,
  nsims = 1,
  tergmLite = TRUE,
  tracked.attributes = c('active', 'status'),
  cumulative.edgelist = TRUE,
  truncate.el.cuml = Inf,
  save.cumulative.edgelist = TRUE,
  save.run = TRUE,
  verbose = FALSE
)
sim <- netsim(est, param, init, control)

nw <- EpiModel:::make_networkDynamic(sim)
nw <- color_tea(nw, old.var = "status", verbose = FALSE)

# 4. Compute layout + render
slice.par <- list(
  start = 1,
  end = 25,
  interval = 1,
  aggregate.dur = 1,
  rule = "any"
)
compute.animation(nw, slice.par = slice.par)
render.d3movie(
  nw,
  vertex.cex = 0.9,
  vertex.col = "ndtvcol",
  edge.col = "darkgrey",
  vertex.border = "lightgrey",
  displaylabels = FALSE,
  filename = paste0(getwd(), "/movie.html")
)
```

### Bug fixes

- **`saveout.R`**: Fixed naming of `raw.records` list elements. The final naming step referenced `".records"` instead of `"raw.records"`, so simulation names (`sim1`, `sim2`, ...) were never assigned to the `raw.records` list in multi-simulation runs.
- **`record_attr_history()`**: Fixed error message that printed `length(unique_ids)` for both "Expected" and "Given", making it impossible to diagnose length mismatches. Now correctly prints `length(value)`.

### Test plan

- [x] `R CMD check --as-cran` passes with no errors or warnings
- [x] `test-tracked-attrs.R`: status change recording in SI model
- [x] `test-tracked-attrs.R`: arrivals and departures with vital dynamics
- [x] `test-tracked-attrs.R`: `get_attr_at` reconstructs correct final state vs saved attributes
- [x] `test-tracked-attrs.R`: input validation errors for out-of-range time/sim
- [x] `test-tracked-attrs.R`: empty `tracked.attributes` produces no history
- [x] `test-tracked-attrs.R`: manual and automatic recording coexist
- [x] `test-newmodules.R`: existing module tests still pass
- [x] Manual: `make_networkDynamic` produces valid `ndtv` animation from `tergmLite` run
- [x] Manual: multi-sim runs have correctly named `raw.records` and `attr.history` elements


smjenness:
Scanned quickly, but didn't you already design something like this before: https://epimodel.github.io/EpiModel/articles/attributes-and-summary-statistics.html#historical-nodal-attributes

AdrienLeGuillou:
It builds on top of it.
The main idea is to automatically check the chosen attributes for changes and store just enough to recreate their value at a given time.
The side effect is that in coordination with cumulative edgelist you can actually recreate a `networkDynamic` from a `tergmLite` simulation and make the `ndtv` animations

smjenness:
## Code Review — `tracked_attrs` vs `main`

*19 files changed · 6 issues found (score ≥ 80)*

---

### `R/net.fn.record.R`

**[Consistency]** Line 72 · Score 92
> `"The size of the \`values\` vector is not equal to the number of nodes selected by the \`posit_ids\` vector nor of length 1."`

Error message still references old parameter names `values` and `posit_ids` after the API was refactored to `value` and `unique_ids`. A user hitting this error would be pointed to the wrong parameter names.

---

### `R/net.fn.track.attrs.R`

**[Bug]** Line 123 · Score 82
> `present_uids <- spells$uid[at >= spells$onset & at < spells$terminus]`

Off-by-one inconsistency with `make_networkDynamic`. `get_attr_at` excludes nodes at their departure time step (`at < terminus`), but `make_networkDynamic` includes them (`terminus + 1` in `activate.vertices`). A node departing at step `t` has `terminus = t` in `get_nodes_spell`. These two functions from the same feature give contradictory answers about node presence at the departure step.

**Fix:** Use `at <= spells$terminus` here, or remove the `+ 1` in `make_networkDynamic` — the key is both functions should agree.

**[Bug]** Line 134 · Score 80
> `dplyr::group_by(uids) |> dplyr::filter(time == max(time)) |>`

`get_attr_at` can return duplicate rows when multiple `attr_history` records exist for the same (attribute, uid, time). The `group_by(uids) + filter(time == max(time))` retains ALL rows at the max time, not just one. If `record_attr_history` is called both automatically and manually for the same attribute+node+timestep, this corrupts the `left_join`.

**Fix:** Use `dplyr::slice_max(time, n = 1, with_ties = FALSE)` instead of the `filter`.

---

### `NEWS.md`

**[Consistency]** Line 7 · Score 90
> `Added get_attr_history() to extract the full attribute history...`

`get_attr_history()` already exists on `main` and is exported. This branch optimized it (split + indexed access), but did not add it. The NEWS entry should describe this as an optimization/enhancement, not a new feature.

**[Consistency]** Line 9 · Score 90
> `Added record_raw_object() for recording arbitrary R objects during a simulation.`

`record_raw_object()` already exists on `main` (confirmed in `R/net.fn.record.R` and NAMESPACE). This is not a new addition in 2.7.0.

---

### `_pkgdown.yml`

**[Consistency]** Line ~85 · Score 85

New exported functions `get_attr_at()` and `make_networkDynamic()` are not listed in `_pkgdown.yml`. Their companion functions (`get_attr_history`, `record_attr_history`) are already listed. Users browsing the pkgdown site won't discover these new functions.

---

**Summary:** 2 bugs · 1 style/convention issue · 3 consistency issues

AdrienLeGuillou:
I fixed the flagged issues.
However, I converted it to draft as I think one or two important changes are required.

1. `make_networkDynamic` should be an internal function and `get_network` should know when to call it (`tergmLite = TRUE` and cumulative edgelist + tracked `active`)
2. actual test to see if we can recreate the `networkDynamic` of a `tergmLite = FALSE` `netsim`. 

I already have the tests but need to implement it in a quick and reliable fashion

AdrienLeGuillou:
### Code review

Found 5 issues:

1. Missing guard for empty attribute changes in `tracked_attrs_record`. Lines 77 and 83 guard with `if (length(...) > 0L)` before calling `record_attr_history`, but line 102 calls it unconditionally, generating spurious empty records when no nodes changed.

https://github.com/EpiModel/EpiModel/blob/7ff0cf5627d9f89393d1170baacd8af3b31542a3/R/net.fn.track.attrs.R#L95-L103

2. `get_attr_at` time validation hardcodes lower bound to 1 instead of using `control$start`. CLAUDE.md says: "`control$start` provides a time offset (default 1). When modifying functions that touch time or epi structure, verify behavior across all three classes." In restart scenarios (`control$start > 1`), querying times below `start` will silently return empty/stale results instead of erroring.

https://github.com/EpiModel/EpiModel/blob/7ff0cf5627d9f89393d1170baacd8af3b31542a3/R/net.fn.track.attrs.R#L137-L138

3. Double call to `update_cumulative_edgelist` at initialization. `initialize.FUN` (via `sim_nets_t1` at `net.mod.simnet.R:219`) already calls `update_cumulative_edgelist` for all networks. The new code at `netsim.R:331-334` calls it again, processing the edgelist twice at step 1. When `truncate.el.cuml > 0`, the second pass could incorrectly terminate edges.

https://github.com/EpiModel/EpiModel/blob/7ff0cf5627d9f89393d1170baacd8af3b31542a3/R/netsim.R#L330-L334

4. Missing separator in `record_attr_history` error message. Line 71 outputs the `at` value and line 72 immediately follows with `"The size of..."`, producing garbled output like `"at time 5The size of the values vector..."`.

https://github.com/EpiModel/EpiModel/blob/7ff0cf5627d9f89393d1170baacd8af3b31542a3/R/net.fn.record.R#L67-L79

5. `DESCRIPTION` version is `2.6.1` but `NEWS.md` introduces `EpiModel 2.7.0`. These should be in sync.

https://github.com/EpiModel/EpiModel/blob/7ff0cf5627d9f89393d1170baacd8af3b31542a3/DESCRIPTION#L1-L3

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

AdrienLeGuillou:
@smjenness ready for your review

smjenness:
Minor note: `update_cumulative_edgelist` is called twice at initialization. The default `initialize.FUN` (`initialize.net`) calls `sim_nets_t1`, which already calls `update_cumulative_edgelist` for all networks (net.mod.simnet.R:217-220). Then `netsim_initialize` calls it again immediately after (netsim.R:331-334). The second call is a no-op (same timestep, unchanged edgelist, no historical edges to truncate), so it's not a bug, just unnecessary work. One of the two call sites could be removed.

smjenness (changes requested):
See comment above

AdrienLeGuillou:
`sim_nets_t1` actually do not call it. The lines mentioned:  217-220 are from `resim_nets`.
I added it to get closer to the "true" networkDynamic that store edges from time = 0. 
As nothing is actually  happening in terms of epidemics, it is reasonable to remove this call for code simplicity and accept that the cumulative edgelist starts at `at == 2`

Issue #990, opened 2026-04-05
set_epi() parameter order inconsistent with other set_*() functions


smjenness:
## Summary

`set_epi()` uses the signature `set_epi(dat, item, at, value)`, inserting the `at` parameter before `value`. All other setter functions in the accessor API use `set_*(dat, item, value)` without an intermediate parameter.

## Details

| Function | Signature | File:Line |
|----------|-----------|-----------|
| `set_param` | `(dat, item, value)` | `R/net.fn.accessor.R:344` |
| `set_control` | `(dat, item, value)` | `R/net.fn.accessor.R:415` |
| `set_init` | `(dat, item, value)` | `R/net.fn.accessor.R:465` |
| `set_attr` | `(dat, item, value, posit_ids, ...)` | `R/net.fn.accessor.R:162` |
| **`set_epi`** | **`(dat, item, at, value)`** | **`R/net.fn.accessor.R:283`** |

The `at` parameter is needed because epi data is time-indexed, but its placement before `value` breaks the consistent `(dat, item, value, ...)` pattern that users learn from the other setters.

## Consequence

Extension developers who learn the API from `set_param(dat, "my.param", 0.5)` will naturally write `set_epi(dat, "i.num", 42, 100)` expecting `(dat, item, value, ...)` — but this actually sets the value `100` at timestep `42`, not value `42` at whatever timestep.

## Discussion

This is a breaking API change to fix, so it may need a deprecation cycle. Alternatively, this could be documented more prominently. The current combined help page for all 31 accessor functions (`@rdname net-accessor`) makes it hard for users to notice this difference.

One option: keep the current signature but add a named-argument convention recommendation in documentation, encouraging `set_epi(dat, "i.num", at = 42, value = 100)`.

PR #1006, opened 2026-04-15
Update `set_epi` signature to match the other accessors conventions

AdrienLeGuillou:
addresses #990 

required for https://github.com/EpiModel/EpiModelHIV-p/tree/v3.4.x


smjenness (changes requested):
## Review

### Bugs

**1. Error message references old parameter name** — `R/net.fn.record.R:72`
```r
"\nThe size of the `values` vector is not equal to the number of nodes ",
```
The parameter was renamed from `values` to `value` in this PR, but the error message still says `` `values` ``. Should be `` `value` ``.

### Typos

**2. "nestim" in test name** — `tests/testthat/test-cumulative_edgelist.R:151`
```r
test_that("nestim can remake the `networkDynamic`", {
```
Should be `"netsim can remake..."`.

**3. Empty bullet in NEWS.md** — `NEWS.md:21`
```markdown
### OTHER

-
```
Bare dash with no content. Either add text or remove the section.

### Minor Code Quality

**4. No-op assignment** — `R/network.R` (in `make_networkDynamic`)
```r
el$start <- el$start  # onset is already correct
```
This is dead code. The comment is useful context, but the assignment does nothing — just the comment would suffice.

### Pre-existing (not introduced by this PR)

**5. Parenthesis bug in `check_attr_lengths`** — `R/net.fn.accessor.R:539`
```r
if (length(wrong_lengths > 0L)) {
```
Should be `if (length(wrong_lengths) > 0L)`. Currently evaluates the comparison inside `length()`, so the condition is always TRUE when `wrong_lengths` has any elements. However, this function is tagged `@keywords internal not_used` and is never called, so it's inert. Not blocking but worth fixing if you're in the area.

PR #1017, opened 2026-04-23
Fix issue 1016 - cumulative edgelists at t=1

### LINKED ISSUE #1016: The optional cumulative edgelist in netsim appears  to drop a few edges
The cumulative edgelist that one can save with `netsim` appears to drop a few edges along the way. See attached MRE:

[cum.edgelist.issue.html](https://github.com/user-attachments/files/26990242/cum.edgelist.issue.html)

 This example only yields dropped edges with onset 0 and terminus 1. However, I first noticed the issue with a more complex set of code embedded within a larger document, and that one also dropped an edge that started later (but was also duration 1) as well as an edge with onset 0 and terminus 2. Most were still onset 0 and terminus 1 though. If it helps, I can attach that code, or it can be run directly from within the Epidemic_Modeling_on_Networks repository.

I checked, and the issue has nothing to do with these dyads appearing twice in the edgelist.

I'm happy to dig even deeper once @smjenness confirms that this isn't the result of me doing something wrong.

smjenness:
Closes #1016.

## Summary

The cumulative edgelist saved by `netsim` (`mod$cumulative.edgelist$sim1` when `cumulative.edgelist = TRUE`) was dropping a few edges relative to the networkDynamic object that `netsim` also saves. In Steve's MRE (attached to #1016), `as.data.frame(get_network(mod, sim = 1))` returned 305 rows while `mod$cumulative.edgelist$sim1` returned 302; the 3 missing spells all had `onset = 0, terminus = 1`. The same bug caused persistent edges to be recorded with `start = 2` instead of `start = 1` (a silent off-by-one), and dropped edges with `onset = 0/terminus = 2` and `onset = 1/terminus = 2` in larger simulations.

## Root cause

`update_cumulative_edgelist()` is called only from `resim_nets()`, which first runs at `at = 2`. Initialization (`sim_nets_t1()` at `at = 1`) never touched the cumulative tracker, so:

1. At `at = 2` the tracker was empty, and persistent edges got `start = 2` (off-by-one).
2. Edges active at `at = 1` but not at `at = 2` (networkDynamic terminus is exclusive, so spells like `onset = 0/terminus = 2` or `onset = 1/terminus = 2` were already gone from `get.dyads.active(net, at = 2)`) were in neither the tracker nor the current edgelist, and silently dropped.
3. Edges active only during the initial ERGM to TERGM step (`onset = 0, terminus = 1`) were dropped for the same reason, and are invisible to `get.dyads.active(net, at = 1)` as well.

## Fix

New internal helper `seed_cumulative_edgelist_t1()` in `R/edgelists.R`, called once at the end of `sim_nets_t1()`:

- Seeds `el_cuml_cur` with edges active at `at = 1` (both tergmLite and non-tergmLite), each with `start = 1`. This resolves the off-by-one and the `onset = 0/terminus = 2` / `onset = 1/terminus = 2` cases. They are now in the tracker at `at = 1` and get properly transitioned to history with `stop = 1` when `update_cumulative_edgelist()` runs at `at = 2`.
- For non-tergmLite (and when `truncate.el.cuml != 0`): additionally seeds `el_cuml_hist` with dyads active at `at = 0` but not at `at = 1`, with `start = 1, stop = 1`. Uses point-in-time `get.dyads.active(nw, at = 0)` / `get.dyads.active(nw, at = 1)` queries (not `as.data.frame(nw)`), to avoid double-counting future spells when `resimulate.network = FALSE` pre-simulates the full trajectory at init.

`update_cumulative_edgelist()` itself is unchanged.

### Design notes

- **`start = 1, stop = 1`** for initial duration-1 edges rather than `stop = 0`: `make_restart_point()` filters `el_cuml_hist` to `stop >= 1`, so `stop = 0` would silently drop these rows on restart.
- **Seed placed inside `sim_nets_t1()`, not `initialize.net()`**: `sim_nets_t1` is only called on the fresh-start branch (`control$start == 1`). On restart (`control$start > 1`), `el_cuml_cur` / `el_cuml_hist` arrive pre-populated from the prior run, so the seed must not fire.
- **tergmLite limitation (documented in NEWS)**: `networkLite` keeps only the current `el` and `lasttoggle`, so initial-step dissolutions are unrecoverable there. The fix seeds what it can (persistent edges at `start = 1`) and leaves the limitation documented.
- **The head/tail semantic reversal** Steve flagged as a side note is preserved; fixing it would be a breaking change for downstream consumers and belongs in a separate issue.

## Files changed

- `R/edgelists.R`: new `seed_cumulative_edgelist_t1()`.
- `R/net.mod.simnet.R`: call the seed at end of `sim_nets_t1()`.
- `tests/testthat/test-cumulative_edgelist.R`: 3 new tests (row-count parity vs networkDynamic, `truncate.el.cuml = 0` hist-seed skip, tergmLite `start = 1`).
- `NEWS.md`: bug-fix entry under 2.6.1 BUG FIXES.

## Test plan

- [x] Added regression test reproducing Steve's MRE in miniature (asserts `nrow(cumulative.edgelist) == nrow(as.data.frame(networkDynamic))` and that initial duration-1 edges appear with `start = 1, stop = 1`).
- [x] Added test for `truncate.el.cuml = 0` to confirm the default active-only behavior is preserved.
- [x] Added tergmLite test confirming the off-by-one fix applies there too.
- [x] Ran full test-cumulative_edgelist.R file locally: 18 pass, 0 fail.
- [x] Reproduced Steve's exact MRE after the fix: cumulative edgelist now has 305 rows (was 302), matching the networkDynamic count; the 3 previously-missing dyads appear with `start = 1, stop = 1`.
- [x] CI: `R CMD check --as-cran` across platforms.


smjenness:
Updated this PR to use option (b) (network-time origin matching `networkDynamic`) following independent review and Steve's preference in #1016. Pushed as 9e16e22.

## What changed vs the original commit

**Old behavior on this branch (option a):**
- Persistent cross-section edges: `start = 1, stop = NA`
- Initial-step dissolutions: `start = 1, stop = 1` (only in non-tergmLite)

**New behavior (option b):**
- Persistent cross-section edges: `start = 0, stop = NA`
- Edges formed during the first TERGM step: `start = 1, stop = NA`
- Edges active in the cross-section but dissolved by the first TERGM step: `start = 0, stop = 0` (mirrors `networkDynamic` spells `[0, 1)`)

The cumulative edgelist now matches `as.data.frame(get_network(sim))` row-for-row **and** value-for-value, including at `at = 0` and `at = 1`. The tergmLite caveat is gone — `sim_nets_t1()` stashes the cross-section edgelist on `dat$run$el_t0_seed` before `simulate_dat()` overwrites `dat$run$el`, so initial-step dissolutions are now recovered there too.

## Why option (b) over option (a)

The original commit fixed the missing-row regression but left an active-set inconsistency at `at = 1`: cross-section-only edges were recorded with `start = stop = 1`, which falsely implied they were active during EpiModel step 1. An independent review (cc'd Codex) found that under option (a) the at=1 active-set count from `cumulative.edgelist` differed from `networkDynamic`'s by exactly the count of `onset=0,terminus=1` spells (Codex's probe: nD=22 vs cum=25). Option (b) eliminates that mismatch and is what users will expect when comparing the two outputs side by side.

Importantly, **EpiModel epidemic outputs (`epi` data frame, `transmat`, prevalence) still begin at `at = 1`** — only the network-time origin in the cumulative edgelist is shifted to align with `networkDynamic`. `infection.net`, `recovery.net`, etc. read the live network via `get.dyads.active(at)`, never the cumulative edgelist, so this PR does not affect epidemic dynamics in any model. The bug was always a post-hoc-output / interpretation issue, not a hidden trajectory artifact.

## Tests

Stronger than #1017's row-count check:

- **Active-set parity** at both `t=0` and `t=1` between `cumulative.edgelist` and `as.data.frame(networkDynamic)`
- **Pointwise spell match**: every `(head, tail, start, stop)` row in the cumulative edgelist maps to a unique `(tail, head, onset, terminus)` row in `networkDynamic` (with `terminus = stop + 1`, `NA` stop ↔ `Inf` terminus, accounting for right-censoring at end of observation)
- **tergmLite parity**: short-duration scenario that forces initial-step dissolutions, asserting they are captured as `start = 0, stop = 0` rows
- `truncate.el.cuml = 0` still skips the hist seed (no regression)

All 22 tests in `test-cumulative_edgelist.R` pass locally.

## Follow-ups (separate issues)

- Network-time vs epidemic-time documentation/vignette (the broader 0-vs-1 question Steve raised) — to be filed as a separate doc issue.
- The pre-existing `head`/`tail` column-naming reversal vs `networkDynamic`'s convention — to be filed as a separate breaking-change candidate.

PR #1025, opened 2026-05-11
Address doc clarity for accessor functions

### LINKED ISSUE #1024: Behavior of `set.control` seems unexpected
See the attached [change_model.html](https://github.com/user-attachments/files/27557967/change_model.html) for full explanation with examples.

TL;DR version: `set_control`  does not work the way I expected it to -- I thought it would take, e.g. a `control_icm` object, not an `icm` object, and return the same. It also appears to strip out the `.FUN` elements. A few questions:

- [ ] Is the observed behavior of stripping out `.FUN` elements expected or a bug?
- [ ] What is the actual use case for `set_control` and its analogs?
- [ ] In a situation where one wants to alter a single element in a `control`, `param` or `init` object, should we be encouraging users to simply edit the object directly? Or is there an accessor function to do this?  And if there isn't one, do we want one?

@smjenness @AdrienLeGuillou

smjenness:
## Summary

Clarifies the intended scope of the `get_*` / `set_*` / `add_*` / `append_*` accessor functions and of `update_params()`, motivated by #1024.

The functions are designed to operate on the live `netsim_dat` object that `netsim()` constructs internally and passes through user-supplied module functions at every time step. They are not a general-purpose tool for editing user-facing `param.net` / `init.net` / `control.net` inputs before a run, nor for editing the object returned by `netsim()`. Applying them outside that context (as in the example in #1024) appears to succeed because everything is list-like, but it produces an object that isn't a valid simulation input. The behavior is intentional and consistent across the whole accessor family, but the docs never said so explicitly.

This PR is documentation-only.

## Changes

- `R/net.fn.accessor.R`
  - One-sentence pointer added to `@description` flagging the intended scope.
  - New `@section Intended Usage:` covering: the expected calling context (inside module functions during a `netsim()` run); what *not* to call these on (input or output objects); and the right alternatives for the common outside-simulation needs (`update_params()`, direct list assignment, fresh constructor calls) and for scheduled mid-simulation changes (`.param.updater.list` / `.control.updater.list`, scenario API).
  - `@seealso` cross-referencing `update_params()`, the input constructors, and `netsim()`.
- `R/net.inputs.R`
  - `update_params()` `@details` extended to contrast outside-simulation use (`update_params`) with inside-module use (`set_param`); notes the lack of `update_controls()` / `update_inits()` siblings and points at `.param.updater.list` / `.control.updater.list` for mid-simulation changes.
  - `@seealso` added.
- `man/net-accessor.Rd` and `man/update_params.Rd` regenerated via `roxygen2::roxygenise()`.

## Test plan

- [x] CI: `R CMD check --as-cran` passes (doc-only change, but confirms cross-references and Rd render).
- [x] Spot-check rendered help: `?\`net-accessor\`` shows the new Intended Usage section; `?update_params` shows the new details paragraph and Seealso block.

Closes #1024 once merged.

### DIFF
--- a/R/net.fn.accessor.R
+++ b/R/net.fn.accessor.R
@@ -3,7 +3,12 @@
 #' @description These `get_`, `set_`, `append_`, and `add_`
 #'              functions allow a safe and efficient way to retrieve and mutate
 #'              the main `netsim_dat` class object of network models
-#'              (typical variable name `dat`).
+#'              (typical variable name `dat`). They are intended for use
+#'              *inside module functions* that run during a [netsim()]
+#'              simulation, not for editing the user-facing `param.net`,
+#'              `init.net`, or `control.net` inputs before a run or the
+#'              output object returned by [netsim()]. See the **Intended
+#'              Usage** section below for details and alternatives.
 #'
 #' @inheritParams recovery.net
 #' @param item A character vector containing the name of the element to access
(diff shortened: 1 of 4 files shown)


sgoodreau:
Text looks good.

PR #1037, opened 2026-05-18
addition of `egor.popsize` to netest

### LINKED ISSUE #936: Consider allowing popsize to be passed as an argument for ergm.ego fits
In ergm.ego there are 2 parameters that influence the size of the network at different points of the estimation process:

* `popsize`  -- sets the scaling of the edges term and is passed as a parameter at the main level.
* `ppopsize` -- controls the size of the pseudo-network used during the ergm estimation and is passed as an argument to `control.ergm.ego`.  It has a helper control `ppopsize.mul`.  The purpose of this parameter is to stabilize estimation and is useful to avoid small network artifacts.

Both parameters have a number of options, the options are partially dependent, and the combinations can be used to represent and produce context-specific outcomes.

As implemented in the `netest` function in `EpiModel` the `popsize` parameter is hardcoded to `0`.  

https://github.com/EpiModel/EpiModel/blob/0256b773c915a1215ea6929495f8ef7190ac9612/R/netest.R#L206-L224

 Per `?ergm.ego` this scales the network edges coefficient to reproduce the size of the sampled network (the number of egos):

Image

By contrast, the default for the `ergm-ego` package is to set `popsize = 1`, which produces the "per-capita" scaling of the edges term:

This hardcoding in `netest` seems unnecessary, and we may want to consider allowing the `popsize` parameter to be passed from `netest`.  In particular, if we allowed `popsize`=1 we would obtain the "per-capita" scaling for the edges term.  That parameterization can be easily used to scale up to a population of arbitrary size, as it is no longer necessary to know the size of the population assumed during estimation.

AdrienLeGuillou:
fixes #936

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -70,3 +70,4 @@ LinkingTo: Rcpp, ergm
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 Config/roxygen2/version: 8.0.0
+RoxygenNote: 7.3.3
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,11 +2,14 @@
 
 ### BREAKING CHANGES
 
+### NEW FEATURES
+
+- Add an `ergm.ego.popsize` argument to `netest()` exposing the `popsize` parameter of `ergm.ego::ergm.ego()`. Defaults to `0` (preserves prior behavior); set to `1` for per-capita scaling of the edges coefficient, which lets the fitted model be applied to networks of arbitrary size. Closes #936.
+
 ### BUG FIXES
 
 ### OTHER
 
-- Add an `egor.popsize` argument to netest to allow scaling the edges term in `ergm.ego` models.
 - Increase `make_module_list` verbosity. It now `message`s if `control$module.order` is not provided and list the modules and there order.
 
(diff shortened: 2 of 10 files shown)


smjenness:
Pushed a few follow-ups:

- Renamed `egor.popsize` to `ergm.ego.popsize` to parallel the existing `set.control.ergm.ego` namespacing.
- Expanded the Roxygen docstring to explain what `0` (default, scales to egor sample size) and `1` (per-capita) mean, and the motivating use case of simulating on networks of different sizes than the egor sample.
- Moved the NEWS entry from `### OTHER` to a new `### NEW FEATURES` section under 2.6.2 and added `Closes #936`.
- Added a test in `test-ergm.ego.R` asserting that `ergm.ego.popsize = 1` shifts the fitted edges coefficient by `-log(N)` relative to the default on the same egor.

Take a look and let me know what you think. Further edits/suggestions are welcome.

smjenness (commented):
Made some further updates for your consideration.

AdrienLeGuillou (commented):
looks good to me.
I tackled it in the smallest possible way and your additions made things clearer

Issue #1030, opened 2026-05-14
Dead nodes not departing


ntparker3:
I have been adapting an EpiModel network model one of my colleagues made (https://github.com/ywang692/PWID_ALIVE_model). She had originally made it on EpiModel 2.4.0 and I had added my modules to that model and had it working on 2.4.0 end of last year. I came back to it recently and now under 2.4.0 it is throwing errors (I’m guessing a dependency updated in the last few months). So I’ve been trying to update it to 2.6.0 (the latest version I get from CRAN) but I am having a lot of issues, particularly with the departure function. I have tried many strategies (deactivate.vertices, depart_nodes, following exactly the tutorial where all you do is set active to 0 and exitTime and let the backend do the rest, etc.) and every time the dead/cessated nodes are not deactivating or losing their edges, meaning new nodes replacing the dead ones are not getting edges when they join because there are no more edges necessary to be added. In the end, all the network stats just tank (see image). I have done test runs where I just keep arrival and/or departure functions and it clearly is the departure function causing this issue.

<img width="1555" height="1022" alt="Image" src="https://github.com/user-attachments/assets/ebb11003-6034-4583-8a1d-b4fb86443f8f" />

I was wondering that since EpiModel is continually being updated, if there might have been changes to the departure function methods that I may have missed. I’m happy to provide more context/code to show what is happening.

smjenness:
Hi @ntparker3 — thanks for the report. A few notes that may point at the cause, with the caveat that I'm working without your actual model code so this is somewhat speculative:

1. **Runtime storage moved under `dat$run` in v2.5.** Starting in EpiModel 2.5 (shipped to CRAN as part of 2.6.0), all runtime simulation state was relocated under `dat$run` — nodal attributes live at `dat$run$attr`, networks at `dat$run$nw`, the unique-id counter at `dat$run$last_unique_id`, etc. This was done to make checkpointing/restart cleaner (saving `dat$run` alone is now sufficient to resume), but it is a breaking change for extension code written against 2.4.0 that touched `dat` directly. See the v2.5 NEWS entry for details.

[...]

PR #1038, opened 2026-05-19
Increase module.order clarity

smjenness:


smjenness:
Hey @AdrienLeGuillou — coming out of #1030, where a user with several custom modules ran into a few `module.order` failure modes that surfaced as cryptic or silent errors. Two changes:

1. **Hard errors at `control.net()` construction** for (a) entries that don't resolve to a supplied `.FUN` (previously surfaced as a cryptic `get_control()` failure deep in `make_module_list()`), and (b) `initialize.FUN` / `verbose.FUN` placed in `module.order` (they run outside the per-step loop, would double-execute).

2. **Warning at `control.net()` construction** if `module.order` omits `resim_nets.FUN`, `summary_nets.FUN`, or `nwupdate.FUN` — the three built-ins whose absence tends to silently break things. Suppressed when the user explicitly disabled one via `NULL`.

**No conflict with `make_module_list()`.** Errors halt before it runs; the warning fires once at construction. `make_module_list()` itself is unchanged, including the #1036 verbose messages. Its `get_control()` error path remains as a safety net for direct post-construction mutation of `ctrl$module.order`.

Docs + tests updated to match (existing fixtures now include the critical built-ins; six new tests cover the validation paths).

Does this look right to you?

AdrienLeGuillou (approved):
That looks right, I followed #1030 from afar and these changes could indeed help.
I have some suggestions for later and potentially breaking changes around these things that we can discuss later.
But for the next release your changes work

Issue #1031, opened 2026-05-16
param.net() table input loses constructor-derived semantics


smjenness:
## Status (updated post-PR #1040)

Most of this issue was resolved by #1040. **Remaining work:** clean up a leftover docstring inconsistency described under "Remaining" below.

- [x] `vital` round-trip bug. Fixed in #1040 (`R/net.inputs.R:259-262` now derives `vital` from both formal args and `p[["..."]]`).
- [x] Regression tests. Added in #1040 (`tests/testthat/test-newmodules.R`, `test_that("param.net preserves data.frame.params values against constructor defaults", ...)`).
- [ ] `data.frame.parameters` alias. **Decision: not added.** Per review (https://github.com/EpiModel/EpiModel/pull/1040#discussion_r3286973147), `data.frame.parameters` was a documentation error, never a previously accepted argument name, so we will not introduce a deprecated alias for it. The canonical (and only) name is `data.frame.params`.

## Remaining

PR #1040 dropped the alias-handling code before merge but kept a docstring edit that still advertises the alias. The roxygen at `R/net.inputs.R:140-141` currently reads:

> `data.frame.params` argument (`data.frame.parameters` is also accepted as a deprecated alias)

The implementation (`R/net.inputs.R:213-216`) only checks for `data.frame.params`. Fix:

- Remove the parenthetical so the doc matches the implementation.
- Re-run `roxygen2::roxygenise()` to regenerate the `man/` page.

Close this issue once that lands.

---

## Original report

The `param.net(data.frame.params = ...)` workflow has several constructor-semantics gaps beyond #1029's `act.rate` overwrite. The table path can preserve raw values but then clobber or misderive values that `param.net()` normally computes from the supplied parameters.

#### `vital` flips to `FALSE` after table round trip

```r
p1 <- param.net(inf.prob = 0.3, a.rate = 0.1, ds.rate = 0.1, di.rate = 0.1)
df <- param.net_to_table(p1)
p2 <- param.net(data.frame.params = df)

p1$vital
#> TRUE
p2$vital
#> FALSE
```

This happens because `param.net()` derives `vital` from `missing(a.rate)`, `missing(ds.rate)`, etc., rather than checking whether those parameters are already present in `p` from the data-frame loader.

#### Documented alias mismatch

[...]

PR #1040, opened 2026-05-20
fix param.net

### LINKED ISSUE #1029: param.net() default for act.rate silently overwrites data.frame.params value
## Summary

`param.net()`'s scalar default for `act.rate` is applied without checking whether `data.frame.params` has already populated `p[["act.rate"]]`, so any user-supplied `act.rate` (or `act.rate_1`, ..., `act.rate_N`) loaded from a `long.param.df` is silently overwritten with `1`. This is the only formal-arg parameter in `param.net()` with this pattern.

## Reproduction

```r
library(EpiModel)

params_df <- data.frame(
  param = c("act.rate_1", "act.rate_2", "act.rate_3", "act.rate_4"),
  value = c(5, 1, 1, 1),
  type  = rep("numeric", 4)
)

# Confirm param.net_from_table() unflattens correctly:
param.net_from_table(params_df)$act.rate
#> [1] 5 1 1 1   <- correct

# But when fed through param.net():
p <- param.net(data.frame.params = params_df)
p$act.rate
#> [1] 1         <- silently clobbered to scalar
```

## Mechanism

In `R/net.inputs.R`, `param.net()` first reads `data.frame.params` and populates `p` (lines 213–219):

```r
if ("data.frame.params" %in% names.dot.args) {
  p <- param.net_from_table(dot.args[["data.frame.params"]])
  ...
}
```

At this point `p[["act.rate"]]` is correctly the unflattened vector. Then at lines 254–256:

```r
if (missing(act.rate)) {
  p[["act.rate"]] <- 1
}
```

`missing(act.rate)` only tests whether `act.rate` was passed as a *named arg to `param.net()`*. It does not check whether `p[["act.rate"]]` has already been set by the data-frame path, so the table-supplied value is unconditionally overwritten.

I audited every other `missing()` and default-write in `param.net()`: `act.rate` is the only formal-arg parameter with this unguarded pattern. The other defaults either set a flag (`vital` at L257–258) or already guard on `is.null(p[[...]])` (`inter.start` at L264–266). That's why every other `_X` parameter in a `long.param.df` (`inf.prob_1..._4`, etc.) unflattens correctly through `param.net()` — only `act.rate` hits the trap.

## Impact

For consumers that index `act.rate` by network layer (e.g., `act.rate[layer]` in multi-layer model code), the scalar overwrite produces `NA` for `layer >= 2`. Those `NA`s propagate through downstream multiplications and silently break the per-edge transmission draw on non-layer-1 networks. So this is not "wrong number, right shape" — it's a real correctness break for any multi-layer model parameterized via `data.frame.params`.

This blocks the `India_COVID19_model` parameter migration from inline `param.net()` to `model_parameters.csv` (EpiModel/India_COVID19_model#37), and would block any other project trying to drive a multi-layer model from a flat CSV.

## Proposed fix

One-line guard on the default, consistent with the `inter.start` guard immediately below it:

```diff
- if (missing(act.rate)) {
+ if (missing(act.rate) && is.null(p[["act.rate"]])) {
    p[["act.rate"]] <- 1
  }
```

This preserves the default for `param.net(inf.prob = 0.3)`-style calls (no `data.frame.params`, scalar `1` fills in) and respects table-supplied values.

## Notes

- A regression test should cover the `data.frame.params` path with `act.rate_X` rows — exercising the unflatten/preserve-through-param.net pipeline, not just `param.net_from_table()` in isolation.
- Worth a quick scan of `control.net()` and `init.net()` for analogous unguarded `missing()` defaults that would have the same interaction with their respective table loaders, since this pattern was only audited within `param.net()`.

AdrienLeGuillou:
fix #1029 & #1031

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 -   Fix `list_special_params` so `param.net_from_table` correctly fails if a reserved parameter name is passed in the parameter table.
 -   Fix the `cumulative.edgelist` recording of the head and tail of nodes. Previously it was inverted. This had no effect on undirected networks but would on directed ones
--   Fix `param.net` issues where `data.frame.params` where overriden by the constructor semantics
+-   Fix `param.net` so that values supplied via `data.frame.params` (notably `act.rate`, and the vital-dynamics parameters `a.rate` / `ds.rate` / `di.rate` / `dr.rate`) are no longer overwritten by the constructor's defaults. Also accept `data.frame.parameters` as a deprecated alias for `data.frame.params` (with a warning) to match the documented argument name. Closes #1029 and #1031.
 
 ### OTHER
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,7 @@
 
 -   Fix `list_special_params` so `param.net_from_table` correctly fails if a reserved parameter name is passed in the parameter table.
 -   Fix the `cumulative.edgelist` recording of the head and tail of nodes. Previously it was inverted. This had no effect on undirected networks but would on directed ones
+-   Fix `param.net` issues where `data.frame.params` where overriden by the constructor semantics
 
 ### OTHER
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 -   Fix `list_special_params` so `param.net_from_table` correctly fails if a reserved parameter name is passed in the parameter table.
 -   Fix the `cumulative.edgelist` recording of the head and tail of nodes. Previously it was inverted. This had no effect on undirected networks but would on directed ones
--   Fix `param.net` so that values supplied via `data.frame.params` (notably `act.rate`, and the vital-dynamics parameters `a.rate` / `ds.rate` / `di.rate` / `dr.rate`) are no longer overwritten by the constructor's defaults. Also accept `data.frame.parameters` as a deprecated alias for `data.frame.params` (with a warning) to match the documented argument name. Closes #1029 and #1031.
+-   Fix `param.net` so that values supplied via `data.frame.params` (notably `act.rate`, and the vital-dynamics parameters `a.rate` / `ds.rate` / `di.rate` / `dr.rate`) are no longer overwritten by the constructor's defaults. Closes #1029 and #1031.
 
 ### OTHER
 
--- a/R/net.inputs.R
+++ b/R/net.inputs.R
@@ -251,11 +251,14 @@ param.net <- function(inf.prob, inter.eff, inter.start, act.rate, rec.rate,
     stop("Parameters using the .m2 suffix have been removed. ",
          "Use the .g2 suffix instead (e.g., inf.prob.g2).")
   }
-  if (missing(act.rate)) {
+  if (missing(act.rate) && is.null(p[["act.rate"]])) {
     p[["act.rate"]] <- 1
   }
-  p[["vital"]] <- ifelse(!missing(a.rate) | !missing(ds.rate) |
-                           !missing(di.rate) | !missing(dr.rate), TRUE, FALSE)
+
+  p[["vital"]] <- !missing(a.rate) | !is.null(p[["a.rate"]]) |
+    !missing(ds.rate) | !is.null(p[["ds.rate"]]) |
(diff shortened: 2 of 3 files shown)


smjenness:
A few things further addressed in the latest commits.

- **Missing regression tests.** Both fixes need coverage — a `data.frame.params` with `act.rate_1..._N` rows asserting vector preservation (#1029), and a `param.net_to_table()` → `param.net()` round-trip asserting `vital` survives (#1031). Without these, this can silently regress.
- **`data.frame.parameters` alias from #1031 isn't addressed.** Docs still reference that name, and the existing test at `tests/testthat/test-newmodules.R:514` calls `param.net(data.frame.parameters = ...)` — which falls through to `...` and never triggers unpacking, so it only checks the class. Suggest accepting it as a deprecated alias with a warning (friendlier for v2.6.2) and fixing that test to use `data.frame.params` and assert the values actually unpack.
- **NEWS.md typo.** `where data.frame.params where overriden` → `were overridden`; also worth naming which params are affected (`act.rate`, `a.rate`, `ds.rate`, `di.rate`, `dr.rate`).

smjenness (commented):
Please consider the changes above.

R/net.inputs.R
```diff
@@ -209,6 +210,17 @@ param.net <- function(inf.prob, inter.eff, inter.start, act.rate, rec.rate,
...
+  # implemented; accept it as a deprecated alias for `data.frame.params`.
+  if ("data.frame.parameters" %in% names.dot.args &&
+        !"data.frame.params" %in% names.dot.args) {
+    warning("`data.frame.parameters` is a deprecated alias for ",
+            "`data.frame.params`; please use `data.frame.params`.")
+    dot.args[["data.frame.params"]] <- dot.args[["data.frame.parameters"]]
+    dot.args[["data.frame.parameters"]] <- NULL
+    names.dot.args <- names(dot.args)
+  }
+
```
AdrienLeGuillou: `data.frame.parameters` was never an alias, just a typo in the doc. 
Unless anyone ever had an issue with it I don't think testing for names is relevant here.

If we fear silent errors, we should ask messages telling saying that a `data.frame.params` was passed and use if it is, and saying that none was passed otherwise.

tests/testthat/test-newmodules.R
```diff
@@ -511,9 +511,22 @@ test_that("Load parameters from data.frame", {
   expect_s3_class(param, "param.net")
   expect_type(param, "list")
 
-  expect_silent(param <- param.net(data.frame.parameters = params.df))
+  expect_silent(param <- param.net(data.frame.params = params.df))
```
AdrienLeGuillou, line 514: as mentioned above, this may not be silent

tests/testthat/test-newmodules.R
```diff
@@ -511,9 +511,22 @@ test_that("Load parameters from data.frame", {
...
   expect_type(param, "list")
+  # Verify the table was actually unpacked (not just stored as a dot arg)
+  expect_equal(param$p1, 10)
+  expect_true(param$p2)
+  expect_equal(param$p3, c(1, 3))
+  expect_equal(param$p4, "tsa")
+
+  # `data.frame.parameters` is accepted as a deprecated alias and emits a warning
+  expect_warning(
+    param_alias <- param.net(data.frame.parameters = params.df),
```
AdrienLeGuillou: see first comment on implementation

NEWS.md
```diff
@@ -10,6 +10,7 @@
 
 -   Fix `list_special_params` so `param.net_from_table` correctly fails if a reserved parameter name is passed in the parameter table.
 -   Fix the `cumulative.edgelist` recording of the head and tail of nodes. Previously it was inverted. This had no effect on undirected networks but would on directed ones
+-   Fix `param.net` so that values supplied via `data.frame.params` (notably `act.rate`, and the vital-dynamics parameters `a.rate` / `ds.rate` / `di.rate` / `dr.rate`) are no longer overwritten by the constructor's defaults. Also accept `data.frame.parameters` as a deprecated alias for `data.frame.params` (with a warning) to match the documented argument name. Closes #1029 and #1031.
```
AdrienLeGuillou: we should not accept `data.frame.parameters`. 
it was a documentation error not a value accepted at anypoint in the past

PR #1052, opened 2026-07-24
Add edist ERGM term for spatial network models

### LINKED ISSUE #60: Develop a dynamic spatial model example
Moved from EpiModel/sismid#82, which is a better home for a Gallery example. Original proposal from @martinamorris, with discussion from @sgoodreau.

---

This is such a common use case for our students, esp those coming from EEB.

Maybe:
* treat spatial location as an attribute
* fit model with both a spatial effect (as a simple distance term) and other effects
* requires a module that implements the spatial migration process (typical example: seasonal cycle)

Maybe extension:
* Consider adding an environmental infection term to represent variable environmental parasite load, for parasitic infections that have more complex life cycle in and out of hosts

Need to consider: What are the typical forms of data in this setting? Grid sampling? Sensor data?

---

From @sgoodreau:

Excellent idea. I believe it ties into something we've needed to do for a very long time: code up a specific `ergm` term for Euclidean distance given two attributes that represent location in 2D space. As far as I understand, it isn't easily doable without that. Unless there is a way to use the new functionality in `ergm` 4.0 to specify this specific function of the attributes on the fly? I know that's a general type of functionality in there but not enough about the specifics. Even if that does work, though, it might be nice to have this term coded up explicitly.

smjenness:
Adds a custom `edist` ERGM term for spatial network models, motivated by EpiModel-Gallery#60. It gives a memory-cheap way to make tie probability depend on Euclidean distance between nodes without constructing and carrying an n-by-n `edgecov` distance matrix.

## What it does

For an edge between nodes i and j with two or more numeric coordinate attributes, `edist` contributes the Euclidean distance `d_ij = sqrt(sum_k (x_ik - x_jk)^2)`, optionally raised to one or more powers.

```r
~edges + edist(c("lat", "long"))              # Euclidean distance
~edges + edist(c("lat", "long"), pow = 2)     # squared Euclidean distance
~edges + edist(c("lat", "long"), pow = c(1, 2))  # polynomial in distance
```

The `pow` argument controls whether the log-odds of a tie are linear or non-linear in distance:

- `pow = 1` (default): log-odds linear in Euclidean distance.
- `pow = 2`: squared Euclidean distance, a Gaussian-type decay in tie probability.
- larger powers: sharper declines in tie probability as distance increases.
- a vector of powers: one statistic per power, so a single term fits a polynomial in distance.

## Why not operators or edgecov

The `Sum`/`Prod` operator route reaches squared Euclidean distance (`Sum(list(~absdiff("lat", pow = 2), ~absdiff("long", pow = 2)), ...)`) but cannot reach true Euclidean distance. Those operators compose network summary statistics, i.e. values already summed over edges, so applying a square root there gives `sqrt(sum_edges d^2)` rather than `sum_edges d`. The square root has to be applied per dyad, inside the change statistic, before summing. `edgecov` with a precomputed distance matrix can do it but costs O(n^2) memory. `edist` computes the distance on the fly from the length-n coordinate vectors, so memory is O(n).

## Implementation

Follows the existing `absdiffby` pattern:

- `src/changestats.users.c`: `d_edist` change statistic (dyad-independent; computes squared distance once per toggle, then raises to each requested power).
- `src/changestats.users.h`: declaration.
- `R/ErgmTerms.R`: `InitErgmTerm.edist`, using `ergm_get_vattr(..., multiple = "matrix")` to gather the coordinate columns and packing `c(ndim, npow, powers, coords)` into `inputs`. Requires two or more dimensions and errors otherwise, pointing single-dimension users to `absdiff`.
- Docs, `_pkgdown.yml` reference entry, and a NEWS item.

## Testing

Added to `tests/testthat/test-terms.R`:

- summary statistics match manual computation for `pow = 1`, `pow = 2`, `pow = 1.5`, and vector `pow`, including coefficient names.
- change statistics under `simulate(..., monitor = )` match manual computation (exercises the add/remove toggle path, not just the empty-to-full path).
- the two-dimension minimum errors as expected.
- three-dimensional coordinates work.

All 27 tests in the file pass locally, including the existing `absdiffby`, `absdiffnodemix`, and `fuzzynodematch` tests. A static `ergm` fit plus simulation on a 200-node network recovers the target edge and distance statistics.

Closes EpiModel/EpiModel-Gallery#60.


sgoodreau (approved):
I reviewed the R code, documentation, and `thattest` tests, and also ran my own tests.  All of it looks great.

smjenness:
Opened statnet/ergm#659 to add this term directly in ergm, where it fits better as a general spatial ERGM term. Keeping this PR open for now and cross-referencing; will revisit once the ergm PR lands.

smjenness:
Closing, better addressed in https://github.com/statnet/ergm/pull/660

PR #1067, opened 2026-08-04
Fix nodal attribute length invariant in set_attr() and append_attr()

### LINKED ISSUE #1058: set_attr() and append_attr() can create attributes that violate the length invariant

Both accessors can create a nodal attribute whose length does not match the number of nodes, which breaks the invariant the rest of the simulation relies on.

## set_attr() with posit_ids on an attribute that does not exist yet

`set_attr()` snapshots the attribute list first, then creates the missing attribute, then writes into the stale snapshot:

```r
set_attr <- function(dat, item, value, posit_ids = NULL, override.length.check = FALSE) {
  attr_list <- raw_get_attr_list(dat)   # snapshot taken here
  if (!item %in% names(attr_list)) {
    dat <- add_attr(dat, item)          # creates a full-length NA vector in dat
  }
  ...
    attr <- attr_list[[item]]           # still NULL: reads the pre-creation snapshot
    attr[posit_ids] <- value
```

`add_attr()` does the right thing, but its result is discarded. Assigning into `NULL` produces a vector of length `max(posit_ids)`, so the resulting length depends on which nodes were selected. In a 10 node model:

| call | resulting length | expected |
|---|---|---|
| `set_attr(dat, "new", c(5, 6), posit_ids = c(1, 2))` | 2 | 10 |
| `set_attr(dat, "new", c(5, 6), posit_ids = c(4, 7))` | 7 | 10 |

Setting `posit_ids` on an attribute that already exists is correct (length 10), so this is specific to the create-then-write path.

## append_attr() on an attribute that does not exist yet

```r
old_vals <- get_attr(dat, item, override.null.error = TRUE)   # NULL
dat <- set_attr(dat, item, c(old_vals, new_vals), override.length.check = TRUE)
```

With `old_vals` NULL, the new attribute is just `new_vals`. In a 10 node model, after `append_core_attr(dat, at, 2)` brings the node count to 12, `append_attr(dat, "brandnew", 99, 2)` produces a length 2 attribute rather than length 12 with `NA` for the 10 pre-existing nodes.

This is easy to hit in an arrivals module that appends a custom attribute on the first time step any arrival occurs, since the attribute does not exist before then.

## check_attr_lengths() does not work and is never called

While confirming the above, `check_attr_lengths()` turns out to be broken in three ways:

```r
attr_lengths <- vapply(attr_list, length, integer(1))
expected_length <- attr_lengths$active            # 1. $ on an atomic vector
wrong_lengths <- which(attr_lengths != expected_length)
if (length(wrong_lengths > 0L)) {                 # 2. paren in the wrong place
  ...
  msg <- c(msg, "`", names(wrong_lengths)[i], "`: ", wrong_lengths[i], "\n")  # 3. index, not length
```

1. `vapply` returns an atomic named integer vector, so `attr_lengths$active` errors with "$ operator is invalid for atomic vectors". The function always errors before checking anything.
2. `length(wrong_lengths > 0L)` should be `length(wrong_lengths) > 0L`. It happens to behave correctly, but only by accident.
3. `wrong_lengths[i]` is the position from `which()`, not the offending length, so the message would report the wrong number.

The function is documented in `man/check_attr_lengths.Rd` but is not in `NAMESPACE` and is not called anywhere in `R/` or `tests/`. It should either be fixed and wired in, or removed.

## Suggested work

- Use the post-creation attribute in `set_attr()`, so a newly created attribute is full length with `NA` for unselected nodes.
- Seed `old_vals` in `append_attr()` with a full-length `NA` vector when the attribute is missing.
- Fix `check_attr_lengths()`, decide whether to export it, and call it somewhere meaningful.
- Tests for selected-node creation, `NA` for unselected and pre-existing nodes, append after `append_core_attr()`, and `check_attr_lengths()` passing afterward.

smjenness:
Closes #1058.

Three fixes, all in the nodal attribute accessors.

`set_attr()` with `posit_ids` on an attribute that does not exist yet snapshotted the attribute list, created the attribute, then wrote into the stale snapshot. Assigning into `NULL` gave a vector as long as the highest selected position, so the length depended on which nodes were picked: length 2 for `posit_ids = c(1, 2)`, length 7 for `c(4, 7)`, in a 10 node model.

`append_attr()` on an attribute that does not exist yet started from `NULL`, so the result held only the appended values. An arrivals module that creates a custom attribute the first time any arrival occurs left that attribute covering the new nodes alone. Both now fill the nodes they do not write to with `NA`.

`check_attr_lengths()` errored on its first line for every input (`$` on the atomic vector `vapply` returns) and reported attribute positions rather than lengths. It is fixed and wired in: `netsim()` calls it at the end of each time step, after the modules and trackers. That is the one behavior change worth watching, since a model that ends a step with a wrong-length nodal attribute now gets a hard error naming the attribute instead of continuing. Temporary inconsistencies within a step are still fine, since the check runs once per step rather than per module. The full local suite is clean, including `test-net-long.R`, so no built-in path trips it.

Tests: creation through `posit_ids`, `NA` for unselected and pre-existing nodes, type preservation through the padding, `append_core_attr()` from an empty attribute list, and `check_attr_lengths()` on both a good and a corrupted object. Also a `netsim()` regression test with an arrivals module that first appends a custom attribute mid-simulation, asserting every attribute covers every node at the end.


R/net.fn.accessor.R
```diff
@@ -257,6 +269,14 @@ append_attr <- function(dat, item, value, n.new) {
   }
 
   old_vals <- get_attr(dat, item, override.null.error = TRUE)
+  if (is.null(old_vals)) {
+    # the attribute does not exist yet: the nodes that predate the ones being
+    # appended get NA, so that the result is still one value per node.
+    # `append_core_attr` runs first, so `active` already covers the new nodes.
+    # At initialization `active` is itself being created, so there is no padding
+    n.active <- length(get_attr(dat, "active", override.null.error = TRUE))
```
AdrienLeGuillou: - convention: `n_active` instead of `n.active` for variables used within the function.
- use `n_active = length(get_posit_ids(dat))` instead. It will fail, as expected, if active is not available

AdrienLeGuillou (changes requested):
One small change requested.

As a more general comment. We may want at some point review the accessors and the attribute creation overall. See #1074 (longer term)

R/net.fn.accessor.R
smjenness: Done in 283dff1c. Renamed to `n_active` and switched to `length(get_posit_ids(dat))`, so a missing `active` now errors instead of being read as zero nodes.

One exception: `active` itself. `append_core_attr()` creates it first, so on the very first call there is no `active` to count and `get_posit_ids()` would error out of every initialization. The padding branch is now skipped when `item == "active"`, which gives the same result as before (nothing to pad). Added a test that appending any other attribute before the nodes exist is an error.

Full check clean under R 4.6.1 with `NOT_CRAN=true`, so the long tests ran too.

PR #1068, opened 2026-08-04
Fix interval merging in dedup_cumulative_edgelist()

### LINKED ISSUE #1063: dedup_cumulative_edgelist() does not merge overlapping spells, and drops open-ended ones

`dedup_cumulative_edgelist()` (`R/reachable.R`) does not produce the merged spells its documentation describes.

## Overlapping spells are not merged

```r
el <- data.frame(head = c(1, 1), tail = c(2, 2), start = c(1, 2), stop = c(3, 4))
EpiModel:::dedup_cumulative_edgelist(el)
#>    head  tail start  stop
#> 1     1     2     1     3
#> 2     1     2     1     4
```

For one dyad with spells `[1, 3]` and `[2, 4]`, the expected result is a single row `[1, 4]`. Instead two rows come back, and the second has had its `start` rewritten to 1 while keeping `stop = 4`, so the output is neither the input nor a correct merge.

Disjoint spells behave correctly:

```r
el2 <- data.frame(head = c(1, 1), tail = c(2, 2), start = c(1, 5), stop = c(2, 6))
EpiModel:::dedup_cumulative_edgelist(el2)
#> two rows, [1, 2] and [5, 6]   <- correct
```

## Open-ended spells are silently dropped

This is the more damaging case. An `NA` stop means the edge is still active:

```r
el3 <- data.frame(head = c(1, 1), tail = c(2, 2), start = c(1, 2), stop = c(3, NA))
EpiModel:::dedup_cumulative_edgelist(el3)
#>    head  tail start  stop
#> 1     1     2     1     3
```

The open-ended spell disappears and the dyad is reported as having ended at time 3. Any reachability or partnership-duration analysis downstream would treat an active partnership as closed.

## Suggested work

Replace the implementation with a proper interval merge over a cumulative end point, per dyad. Tests for:

- simple overlaps;
- transitive overlap chains (`[1, 3]`, `[2, 5]`, `[4, 8]` collapsing to one row);
- disjoint spells;
- open-ended (`NA`) stops, including an `NA` spell that overlaps a closed one;
- multiple dyads in one edgelist.

The `NA` semantics should be settled first: an `NA` stop overlapping an earlier closed spell should presumably absorb it and stay open-ended.

smjenness:
Closes #1063.

The function did not merge the spells it is documented to merge. For an edge with spells `[1, 3]` and `[2, 4]` it returned two rows, the second with its `start` rewritten to 1 and its `stop` left at 4, so the output was neither the input nor a correct merge. The `max()` / `min()` calls ran over the whole group rather than the pair being compared, and the lag comparison only looked one row back, so overlap chains never collapsed.

The `NA` stop case was worse: an `NA` stop marks an edge never observed to dissolve, and the overlap test required both bounds to be non-missing, so the open-ended spell fell through the merge and was dropped by the closing `unique()`. An active partnership came back reported as ended.

Replaced with an interval merge over a running end point, per edge. Spells are sorted by start, `NA` stops act as `Inf`, and any spell starting at or before the running end point is absorbed. So overlaps collapse transitively, an open-ended spell absorbs everything starting after it and stays open-ended, and spells that merely touch stay separate since they do not overlap.

Two smaller decisions worth a look:

- Spells are grouped by every column other than `start` and `stop`. For the plain four-column edgelist that is the same as grouping by `(head, tail)`, but it also means a `network` column (as `get_cumulative_edgelists_df()` returns) merges within each network rather than across them.
- A missing `start` is now an error rather than silently producing nonsense. Output is sorted by edge then `start`, and storage modes are preserved.

Tests cover simple overlaps, containment, transitive chains, disjoint and merely-touching spells, unsorted input, all the `NA` stop combinations, multiple edges, the extra-column grouping, empty input, class and type preservation, and the `NA` start error. The roxygen example is now runnable rather than `\dontrun`.


AdrienLeGuillou (commented):
Is this function actually used by any project? 

Can you provide actual cases where the function failed? From `netsim` or `networkDynamic` cumulative edgelists

PR #1072, opened 2026-08-04
Handle every per-simulation slot consistently in netsim objects

### LINKED ISSUE #1059: Per-simulation output slots misalign in saveout, get_sims(), and merge.netsim()

Several per-simulation slots on a `netsim` object are not handled consistently, so an object's `control$nsims` can disagree with the per-simulation state it actually carries.

Reproduced with a 3 simulation run using `cumulative.edgelist = TRUE`, `save.cumulative.edgelist = TRUE`, `save.network = TRUE`, and a module calling `record_raw_object()`.

## raw.records is never named

`saveout.net` writes the slot as `raw.records`:

```r
out$raw.records[s] <- list(dat$raw.records)   # R/saveout.R:225
```

but the naming loop names `".records"`, which does not exist:

```r
top_lvl_elts <- c("attr.history", ".records")  # R/saveout.R:300
```

`name_saveout_elts()` is given `NULL`, matches neither branch, and returns silently. So `attr.history` gets `sim1`, `sim2`, ... and `raw.records` stays unnamed. Observed: `raw.records` has length 3 and no names.

## get_sims() leaves several slots unsubset

`get_sims()` subsets `epi`, `run`, `network`, `stats$nwstats`, `stats$transmat`, `diss.stats`, and `control$save.other`, but not `coef.form`, `cumulative.edgelist`, `attr.history`, or `raw.records`.

`get_sims(sim, 2)` on the 3 simulation object:

| slot | length after | names after |
|---|---|---|
| `control$nsims` | 1 | |
| `network` | 1 | `sim1` |
| `coef.form` | 3 | `sim1, sim2, sim3` |
| `cumulative.edgelist` | 3 | `sim1, sim2, sim3` |
| `raw.records` | 3 | none |

Beyond the length mismatch, the retained `coef.form$sim1` is the coefficient set from the *original* simulation 1, not from the simulation that was kept. If that subset is later used as a restart point, it selects the wrong formation coefficients.

## merge.netsim() keeps only the first object's copies

`z <- x` and then only `epi`, `transmat`, `network`, `nwstats`, `save.other`, and `diss.stats` are merged. `run`, `coef.form`, `cumulative.edgelist`, `attr.history`, and `raw.records` silently retain `x`'s values.

Merging two 3 simulation objects:

| slot | length after |
|---|---|
| `control$nsims` | 6 |
| `epi$i.num` columns | 6 |
| `network` | 6 |
| `coef.form` | 3 |
| `cumulative.edgelist` | 3 |
| `raw.records` | 3 |

## Suggested work

Handle these slots together rather than one at a time, since the three code paths need to agree on what counts as a per-simulation slot. Tests for subsetting, merging, naming, and using a subset as a restart point.

This is separate from #724, whose original random-parameter scope was resolved in v2.6.1.

smjenness:
Closes #1059.

The three code paths that touch per-simulation state each worked from their own list of what counts as per-simulation. They now share one, `netsim_sim_slots()` in `R/saveout.R`, which returns the paths into a `netsim` object holding one element per simulation and drops the ones a given object does not carry. Paths are character vectors, so recursive `[[` reaches both the top-level slots and the ones under `stats`. `epi` stays separate, since its simulations are data frame columns, and `param$random.params.values` keeps its existing helpers.

That fixes the three symptoms in the issue: `raw.records` is named (`saveout.net()` was naming `".records"`), `get_sims()` subsets `coef.form`, `cumulative.edgelist`, `attr.history`, and `raw.records` along with the rest, and `merge.netsim()` merges them rather than keeping only `x`'s.

The `coef.form` case was a wrong answer rather than a length mismatch. A restart reads `run[[s]]` and `coef.form[[s]]` positionally, so restarting from `get_sims(x, 2)` used simulation 2's network state with simulation 1's formation coefficients. There is a test for it now, built on a model with vital dynamics, where the per-simulation coefficients genuinely drift apart (-6.3796, -6.4734, -6.3796 in my run), so it would have caught the old behavior.

Two follow-on changes in `get_sims()`, both flagged in the plan:

- Simulations come back in the order given in `sims`. The old code deleted the complement, so `sims = c(3, 1)` silently returned 1 then 3.
- The `epi` columns are renumbered along with everything else. Before, a one-simulation subset had `run$sim1` sitting beside an `epi` column still named `sim2`. `as.data.frame()` indexes `epi` positionally, so nothing downstream depended on the old names.

One extra fix, directly in the way of the split-and-merge workflow you described: `merge.netsim()` compared `set.control.ergm` and `set.control.tergm` with `identical()`, and those `statnet` control objects carry formulas whose environments differ between two calls to `control.net()`. So merging two independently parameterized runs failed with "x and y have different controls" unless the very same control object had been reused, which defeats the manual parallelization the function exists for. They are now compared by value, matching what `monitors` and `nwstats.formula` already did. I hit this writing the merge tests; say the word if you would rather it were split into its own PR.

Tests, on a fixture carrying every slot (saved run state, networks, nwstats, transmat, diss.stats, cumulative edgelists, a `save.other` element, and both record types): naming straight out of `netsim()`; `get_sims()` length, names, and content per slot, plus ordering, class preservation, and the `sims = 0` guard; `merge()` across two runs and the split-then-merge round trip asserting slot-by-slot equality with the original; the `keep.*` flags still dropping their slots while the unflagged ones still merge; and the restart. Clean runs of `test-get.R` (140), `test-merge.R` (92), `test-netsim.R` (86), `test-truncate-restart.R` (26), `test-netsim-parallel.R` (15), `test-net-long.R` (141), `test-netattr.R` (81), and `test-newmodules.R` (105).


AdrienLeGuillou (commented):
The code seems to be good but I would like to see full runs with EpiModel HIV (or other) including restart points creation and actual complex simulations.

The objects incriminated are not used often and the risk of breaking things is always high when touching this part of the code.
On this subject, I opened issue #1075 to discuss a more the question of the `netsim` data structure and if it fits our needs

smjenness:
Ok, I will leave this one open and let you test further on EpiModelHIV and explore more via #1075. Also bumping the associated #1059 to v2.7.0 milestone.

PR #1078, opened 2026-09-10
Which sim to use from a restart object

AdrienLeGuillou:
Implement a control to chose which sim object to use from a restart. 
Current implementation match the sim number with recycling. (does it work?)
I want to implement a Random strategy that simply sample for each sim and a fixed strategy to gives a value (for reproducibility)

Also, `make_restart_point` will now be able to create a multi sim object for restart


AdrienLeGuillou:
The features are correctly in place:

1. `make_restart_point` now makes multi-sim objects to restart with while keeping the trimming correct
2. `initialize.net` reads `control$randomize.restart` to decide whether to chose one of the restart point at random or to match the sim numbers with recycling (previous behavior, the default)

I tested with EpiModelHIV that the reported sim object used for restart was actually the correct one by checking epi values and doctored attributes to ensure attribution

AdrienLeGuillou:
This PR now also implements merging more elements with `merge.netsim` as `run`, `attr.history` or `coef.form` where not considered by it

smjenness (commented):
Thanks Adrien, this is the piece we need to spread replicates across a pool of restarts instead of tying every run to one. I went through it against a local build of the branch and have a fair number of suggestions, roughly in priority order. Several are pre-existing, but they get much easier to hit once restart points hold several simulations and merged objects are used as pools.

1. `get_sims()` subsets `run`, `epi` and `stats` but not `coef.form` (or the top-level `cumulative.edgelist`, `attr.history` and `raw.records`), while `initialize.net()` reads `x$coef.form[[s]]` by the same position as `x$run[[s]]`. So `make_restart_point(sim, ..., sims_num = c(1, 3))` pairs sim 3's run with sim 2's `coef.form`, and `sims_num = 2` pairs sim 2's run with sim 1's. These entries differ whenever population size changes, because `edges_correct()` adjusts each simulation's edges coefficient to its own population and `saveout.net()` stores one per simulation, so the restarted network runs at the wrong density for the whole run. In a small open-population test, a restart paired with another simulation's entry ran 16% thinner when the two populations differed by 16%. This predates the PR for `sim_num > 1`, but the PR makes it reachable for any subset, and it is the same kind of bug the NEWS entry fixes in `merge.netsim()`. Subsetting `coef.form` next to `run` in `get_sims()` fixed it in my testing, and a `length(x$coef.form) == length(x$run)` check in `initialize.net()` would guard against older restart files.

2. `merge.netsim()` binds the per-simulation lists (`coef.form`, `run`, `cumulative.edgelist`, `attr.history`, `raw.records`) with `c()` and then assigns `newnames` without checking that each has `nsims` entries. Merging two restart points built from 2-sim outputs gives a `coef.form` of length 4 named `sim1, sim2, NA, NA`, so the second run in the pool is paired with the wrong entry. Restart files made before a fix to item 1 still carry every simulation's entry, so this matters even after that fix. Merging an object that 2.6.2 already merged, which holds only `x`'s `coef.form`, also now fails with a `names` length error. A length check with a clear message would catch both.

3. Merged objects can't be merged again. The default `keep.cumulative.edgelist = FALSE`, the zero-length `attr.history` and `raw.records` that `make_restart_point()` leaves behind, and `keep.run = FALSE` all remove the element from `z`, so the next `merge()` fails the `identical(names(x), names(y))` check with "x and y have different structure". Merging three outputs saved with `save.cumulative.edgelist = TRUE` works on 2.6.2 and errors here, `EpiModelHPC::merge_netsim_scenarios()` merges batches exactly this way without passing the new arguments, and three restart points can't be merged into a pool at all. Defaulting `keep.cumulative.edgelist = TRUE` and leaving zero-length elements in place fixed the default cases for me (one existing test asserts the `FALSE` default). Because `keep.run = FALSE` hits the same removal, a merge loop also can't opt out of keeping every batch's `run`, which is large under the `save.run = TRUE` default in `control_msm()`.

4. The source run is chosen from the simulation number within one `netsim()` call. EpiModelHPC runs every batch as sims 1 to `n_cores`, so the default recycling uses the same first `n_cores` restarts in every batch and never reaches the rest of a larger pool. For paired scenario comparisons we also need batch b of every scenario to use the same restarts. The fixed strategy from your description would cover both: an explicit vector of source indices, one per simulation, that EpiModelHPC or a swfcalib model function can compute from the batch or proposal index.

5. `randomize.restart = TRUE` draws each source independently with replacement. With a pool of 8 and `nsims = 8`, runs used 5.4 distinct restarts on average over 20 seeds (seed 42 drew 1, 5, 1, 1, 2, 4, 2, 2). At 8 replicates that makes the restart share of the variance of the replicate mean about 1.9 times what a balanced assignment gives. A shuffled balanced vector, `sample(rep_len(seq_len(J), nsims))`, drawn once in `netsim()`, would avoid it. Scenarios run with the same seed do get the same draws, on both the sequential and future paths.

6. Making the choice inside `initialize.net()` has a few side effects. The bare `if (control$randomize.restart)` fails with "argument is of length zero" when `initialize.net()` is called directly with a `control.net()` control and `start > 1`, since only `netsim_validate_control()` sets the default. EpiModelHIV defines `reinit_msm <- initialize.net`, which copies the function body when EpiModelHIV is installed, so EpiModelHIV restarts will ignore `randomize.restart` until EpiModelHIV is rebuilt; a wrapper there would avoid that going forward. Deciding the assignment once in `netsim()` would also let each worker receive only its own run. Right now `future_lapply()` ships the whole pool to every worker, twice (once in the closure environment and once as a global), which predates this PR but gets expensive with pools.

7. The chosen source is hard to recover afterward. `_restart_simnum` lives in `dat$run`, which is only saved with `save.run = TRUE` (production runs usually drop it because it is most of the object), and it records the position in the pool rather than the original simulation (`sims_num = c(2, 3)` records 1 and 2). The new verbose message only prints when `nsims == 1`, since `netsim_validate_control()` turns verbose off otherwise. Saving the assignment as a top-level per-simulation element that `merge()` and `get_sims()` handle would let us treat the restart as a random effect in the analysis. `save.other = "_restart_simnum"` works in the meantime.

8. In the restart branch, parameters missing from the new `param` are filled from `x$param`, which `saveout.net()` takes from simulation 1. A parameter drawn through `random.params` and not supplied again is therefore set to sim 1's draw in every restarted simulation, whatever its source (draws of 0.51, 0.20 and 0.23 all restarted at 0.51). Pre-existing, but the new all-simulations default makes it the common case.

9. When `save.other` includes `"run"`, `get_sims()` removes `delsim` from `out$run` twice, once in its `run` block and again in the `save.other` loop, so `make_restart_point()` errors or returns `NULL` runs for any selection other than a leading block. The new merge test uses that configuration. Pre-existing.

10. `merge.netsim()` compares controls with `identical()` and only special-cases `monitors` and `nwstats.formula`, but `set.control.tergm` also holds a formula whose environment differs after `saveRDS()` and `readRDS()`. Merging saved batch files, which is how most pools will be built, fails with "x and y have different controls" unless `param.error = FALSE`. Pre-existing.

11. A few smaller things in `make_restart_point()`:
    - The `nwstats` and `transmat` trims are gated on `save.nwstats` and `save.transmat` rather than on whether those elements exist, so it fails with "subscript out of bounds" on the output of `merge(x, y, keep.nwstats = FALSE)`, which the single-simulation path on main handled.
    - `unique_id` and the cumulative edgelists are re-based by each run's offset, but the kept transmission matrix only has `at` shifted, so with `keep_steps > 1` in an open population its `sus` and `inf` IDs point at the wrong nodes. Pre-existing.
    - `sims_num` accepts duplicates and `integer(0)`: `c(1, 2, 3, 1)` returns `nsims = 4` with 3 runs, `c(2, 2)` fails on `names`, and `c(3, 1)` silently comes back in the order 1, 3.
    - The time attribute check pastes `missing_names` instead of `missing_attrs`, so its error lists nothing.

12. `sim_num` was renamed to `sims_num` with no pointer to the new name, and the default changed from sim 1 to all simulations. Positional calls still work, but named `sim_num =` calls, including the old documented example and some of our project scripts, now fail with "unused argument", and the BREAKING CHANGES entry doesn't mention either change. I'd keep `sim_num` and accept a vector, or error with a pointer the way `control.net()` does for removed arguments. Relatedly, `randomize.restart` should probably be a formal `control.net()` argument with its own `@param`, like `edges.correct.attr`, so a misspelling doesn't pass silently through `...`.

13. The new tests can't fail for most of the above. The merge test uses a closed population, so every simulation's `coef.form` is identical and a wrong binding still passes. `build_restart_sim()` has no demography, so `min(unique_id)` is always 1, and `all(tsmt$at > 0)` holds by construction of the trim. Nothing runs `randomize.restart = TRUE`. A small model with arrivals and departures would make these tests meaningful. Separately, `test-netattr.R` still passes `sim_num = 1` inside an `expect_error()` with no pattern, so it now passes on the argument error instead of the tergmLite check.

14. Docs and NEWS: the new roxygen example calls `sims_num = c(1, 3)` on `sim` right after it is overwritten by a run from a one-simulation restart, so it errors; the details still say `stats` are dropped, though they are kept and trimmed; the `control.net()` section says randomization applies when `nsims` differs from the pool, but it applies regardless; and the entries sit under 2.6.2, which is already tagged. The `merge.netsim()` bug-fix entry also doesn't quite match main, where the merged `run` held only `x`'s runs and those were recycled, so `y`'s runs were silently unused rather than paired with the wrong coefficients.

Items 1 to 4 are the ones I'd want settled before merging. Happy to test the next round.

AdrienLeGuillou:
1. done
2. I don't understand this one. The current way `merge.netsim` works ensure that the `x` and `y` are similar in structure
3. Yes, my workaround was to use copy the control from the first batch and duplicate it to the other ones. This one could be an issue on it's own. My suggestion would be to always do full merge and maybe make another utility to trim `netsim` object. Trying to do both here is very error prone
4. An explicit vector would not fix it here as for the function mentionned `s == 1` always. I do not think it is a relevant issue here. The static restart will probably be a debug case anyway.
5. Sampling without replacement does not incur the same variance and do not have the same statistical properties as sampling with replacement. I need to investigate the consequences of both
6. a. `reinit_msm`: this should be deprecated and `initialized.net` should be the default in the template going forward. 
6. b. restart choice outside of `initialize.net` -> longer term issue. With epimodelHPC default, not and issue as `multicore` is the default parallelization.
7. If one need the `_restart_simnum`, that would be for debugging, `run` would be required as well
8. Bigger picture, what to do with radom params
9.  fixed with better keep semantics
10. see (3.)
11. `make_restart_point` and `merge.netsim` are 2 different functions with different expectation. It's an issue for `merge.netsim`. See (3.)
12. No one was using the function before and there is a message telling that all sims are used

AdrienLeGuillou:
It may be interesting to have a `merge_restart_points` function for now, tailored specifically to the restart points needs
@smjenness  what do you think?

AdrienLeGuillou:
I am revising my position on 7.b., about the fact that the current implementation reads the full `restart_pool` to only chose one.

It's indeed not an issue with `future` `multicore` (de facto default on the HPC), but outside its almost unmanageable. 

There a multiple options to circumvent this, some listed in no particular order:
- multiple restart point files, only one being read by the sim -> implications about parallelization
- choice at top-level netsim, before parallelization. -> implications about the initiation

